PR #2, opened 2018-06-29
port pug logic to vue & add calendar component

ctsexton:
I think that's all the pug logic removed now as far as I can see. I added a partnerLink to the content file as well to check if that's working. Hopefully I got the syntax all right? My first time writing in both pug and vue!

### DIFF
--- a/assets/styles/components/_calendar.scss
+++ b/assets/styles/components/_calendar.scss
@@ -1,0 +1,60 @@
+.cal-title {
+  color: $color-red;
+}
+.cal-container {
+  padding: 5px;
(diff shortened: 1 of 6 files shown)


ctsexton:
Calendar component pulls from hackforla google calendar.

jjandoc:
Thanks for that. I did a bit of refactoring which you can see [here](https://github.com/hackforla/website/commit/9421e24a1e36f3f7f34b43f63b9d7efbe8191b4e). You can see a dev build of the new site at: https://h4la.netlify.com/. Looks like we need to whitelist that domain, though.

PR #11, opened 2018-11-14
Generalize calendar filter

### LINKED ISSUE #10: Filter out strings that mention the hack nights
Right now the code is set up to filter out the calendar items referencing our normal hack nights.  However, because it's looking for an exact match using "!==",  it is allowing the "Canceled" events to still show.  The comparison needs to look for "Westside Hack Night" or "DTLA Hack Night" within the event name string instead of an exact match.

https://github.com/hackforla/website/blob/98d11aca31b0f67ba88bdbc504e44d0eec944e5b/components/Calendar.vue#L41

thekaveman:
This change removes not only regular Hack Night events, but also the canceled events that have been showing up recently.

Closes #10, but I will point out that this change is *slightly* more general than what was called for in that issue. Happy to adjust if needed.

### DIFF
--- a/components/Calendar.vue
+++ b/components/Calendar.vue
@@ -38,10 +38,7 @@ export default {
         )
         .then(({ data }) => {
           this.calendar = data.items.filter(
-            item =>
-              item.summary &&
-              item.summary !== 'Westside Hack Night' &&
-              item.summary !== 'DTLA Hack Night'
+            item => item.summary && item.summary.indexOf('Hack Night') < 0
           );
         });
     },


matikin9:
whoops, not sure what happened @thekaveman but none of the calendar items are showing anymore.

thekaveman:
@matikin9 I didn't do it I swear!!!

It looks like they are back now, I have no idea what happened. Google got some of their traffic "re-routed" the other day, maybe they are experiencing something similar today?

matikin9:
Lol, omg, why can't the internet ever work like it's supposed to!  Anyway,
thanks for the update, glad to see it works now!

thekaveman:
And it's gone again... wtf...

matikin9:
weirrrrd!  Looks okay right now 🤞

PR #24, opened 2018-12-13
Redirect to HTTPS versions of urls

thekaveman:


thekaveman:
Seems to be doing this automatically now that the Let's Encrypt cert is configured.

thekaveman:
It's strange - when I closed this PR without merging it, the automatic HTTPS redirection stopped working. Now that I opened this PR back up, the HTTPS redirection is working again.

Almost as if the updated `_redirects` file had been in play on the production site?

But the other new redirects I added (e.g. `/meetup`) is not working on the production site, so this is very strange...

PR #25, opened 2018-12-18
fixed calendar date issue

andrewyin1994:
It looks like Google Calendar gives datetime in UTC, and the Date object defaults time to midnight. LA is UTC-8 which caused the date to shift to the previous day.

Also, I just got onto this project yesterday, so I'm not sure which branch I should be working on, so apologies if this is the wrong branch.


thekaveman:
Hi @andrewyin1994, sorry about the confusion with the branches. Would you mind rebasing your branch on top of `master` and force-pushing to this PR?

PR #39, opened 2019-03-13
add bourbon & normalize style libraries

abregorivas:
This adds bourbon & normalize libraries to site and connects them to the head.html


abregorivas:
no longer needed

PR #45, opened 2019-03-26
Add Styles and initiate images folder

abregorivas:
The pull request covers the following:
- Adds all base styles
- updates the config.yml file to use sass partials & imports
- Updates the head.html (_includes) to use the generated styles


thekaveman (changes requested):
Remember we had to get rid of the `site/` folder in order for GitHub pages to build (it has to build directly from the root). So I think all your additions just need to shift up by one directory, into the root of the repository.

And then we don't need that modification in `_config.yml` or the `site/` folder.

abregorivas:
I made the following changes.
- Removed sass configuration in config.yml
- moved all files up to root

thekaveman (approved):

thekaveman:
This looks @abregorivas! I'm going to squash merge to get rid of those sync commits, but we'll still have the full history in this PR.

PR #46, opened 2019-03-27
Improve setup docs

thekaveman:
@ExperimentsInHonesty would you mind taking a look at this updated README?


ExperimentsInHonesty:
I think this is good to go for this round.

PR #48, opened 2019-03-27
Header, Nav &  Footer added & hooked to Default.html 

abregorivas:


thekaveman:
@abregorivas it looks like this PR contains both the header and footer... which is totally fine. I'll just close your other one #40 and we'll use this one, sound good?

abregorivas:
Yes Sounds good.

_data/navigation/main.yml
```diff
@@ -0,0 +1,29 @@
...
+  link: about
+  list_class_name:
+  icon_url: /svg/icon-about.svg
+  icon_class: icon icon-hack-nights nav-icon icon icon-projects nav-icon icon icon-about nav-icon
+
+- name: Contact Us
+  link: contact
+  list_class_name: hide-mobile
+  icon_url:
+  icon_class:
```
thekaveman: it looks like each `icon_class` is part of the `.svg` files - any reason to duplicate here? Or any way to pull out of the `.svg` files and generate dynamically from this data?

There's a lot of duplication - `icon nav-icon icon-hack-nights icon-projects` is on almost every one, some multiple times.

_data/navigation/social.yml
```diff
@@ -0,0 +1,35 @@
...
+
+- name: Medium
+  link: https://medium.com/hack-for-la
+  icon_url: /svg/icon-medium
+  header: true
+  footer: true
+
+- name: Twitter
+  link: https://twitter.com/HackForLA
+  icon_url: /svg/icon-twitter
```
thekaveman: I think we should use the full `icon_url` for these - the template shouldn't have to know it needs to add the `.svg` extension to render these.

_includes/footer.html
```diff
@@ -0,0 +1,20 @@
...
+  <div class="page-contain">
+    <nav class="footer-nav">
+      <div class="social-links">
+        <h2 class="social-header">Follow us&mdash;</h2>
+        <ul class="inline-list social-list">
+          {% for item in site.data.navigation.social %} {% if item.footer %}
+          <li>
+            <a href="{{ item.link }}" target="_blank" class="js-social-link js-social-link-{{ item.name | downcase }}">
+              {% if item.icon_url %} {%- include {{ site.baseurl
+              }}{{ item.icon_url }}.svg -%} {% endif %}
```
thekaveman:
```html
{%- include {{ site.baseurl
}}{{ item.icon_url }}.svg -%}
```

I've seen this pattern a few times - any particular reason for the usage of `{%-` ... `-%}` instead of the usual `{%` ... `%}` ? Any reason for that line break?

thekaveman (changes requested):

_data/navigation/main.yml
```diff
@@ -0,0 +1,29 @@
...
+  icon_class: icon icon-hack-nights nav-icon icon icon-projects nav-icon icon icon-press nav-icon
+
+- name: About
+  link: about
+  list_class_name:
+  icon_url: /svg/icon-about.svg
+  icon_class: icon icon-hack-nights nav-icon icon icon-projects nav-icon icon icon-about nav-icon
+
+- name: Contact Us
+  link: contact
```
thekaveman: Each of these `link` should include the `#` symbol at the beginning; since we're still in single-page-website mode for now, they are jump links vs. page links.

thekaveman (changes requested):

_includes/footer.html
abregorivas: my understanding is that the "-" cleans up empty spaces

_includes/footer.html
```diff
@@ -0,0 +1,20 @@
+<footer class="main-footer">
+  <div class="page-contain">
+    <nav class="footer-nav">
+      <div class="social-links">
+        <h2 class="social-header">Follow us&mdash;</h2>
+        <ul class="inline-list social-list">
+          {% for item in site.data.navigation.social %} {% if item.footer %}
```
thekaveman, line 7: You could also do something like:

```
{% assign nav = site.data.navigation.social | where: "footer" %}
{% for item in nav %}
```

So using an `assign` and the `where` filter instead of an `if` block... just an option

thekaveman:
@abregorivas would you mind rebasing on the latest `gh-pages` again?

abregorivas:
changes made and rebase

thekaveman (approved):

thekaveman:
Rebased one last time to deal with conflicts.

PR #52, opened 2019-03-28
Add Sponsors Sections

abregorivas:
The following changes were made:
- Create sponsors _include
- Add all sponsors images
- Set up data file
- Connect to index


_data/sponsors.yml
```diff
@@ -0,0 +1,12 @@
+- name: Code for America
```
thekaveman: I feel like we should treat `sponsors` the same way as `projects` in terms of the data, and they should be a collection. Functionally it won't change the way your code is currently using that data too much, but it allows for some other scenarios (like sponsor-specific pages) as well.

index.html
```diff
@@ -3,21 +3,22 @@
...
               <h4 class="project-title">{{ item.title }}</h4>
               <p class="project-description">{{ item.dek }}</p>
             </div>
+          </div>
         </div>
-      </div>
-    </li>
-    {% endfor %}
+      </li>
+      {% endfor %}
```
thekaveman: This is minor... while these were all just whitespace changes, #50 deals with that by moving it all to an `_include`. We may have some merge conflicts to deal with anyway, so not a huge deal.

thekaveman (changes requested):

thekaveman (approved):

PR #60, opened 2019-04-10
Added Hack Nights Section to Events

ddrosario:
Created a container for all the events.
Added Hack Nights section to the Events container.


DDRosario:
Renamed to included the hyphen. Also changed la-kretz to downtown-la and removed the events wrapper

thekaveman (approved):

PR #70, opened 2019-04-24
Updated image relative url

stu562:
updated base config file to '/website'

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -8,7 +8,7 @@ email: hello@hackforla.org
 description: Hack for LA's website!
 
 baseurl: "/website"
-url: https://hackforla.github.io/website/
+url: https://hackforla.github.io
 
 
 collections:
--- a/_config.yml
+++ b/_config.yml
@@ -7,7 +7,7 @@ title: Hack for LA
 email: hello@hackforla.org
 description: Hack for LA's website!
 
-baseurl: ""
+baseurl: "/website"
 url: https://hackforla.github.io/website/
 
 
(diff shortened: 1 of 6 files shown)


_config.yml
```diff
@@ -7,7 +7,7 @@ title: Hack for LA
 email: hello@hackforla.org
 description: Hack for LA's website!
 
-baseurl: ""
+baseurl: "/website"
 url: https://hackforla.github.io/website/
```
thekaveman: I think we need to remove `/website/` from the `url` property, since it is now on the `baseurl`.

_includes/head.html
```diff
@@ -6,7 +6,7 @@
   <title> {% if page.title %}{{ page.title | escape }} - {% endif %}{{ site.title | escape}}</title>
   <meta name="description"
     content="{% if page.excerpt %}{{ page.excerpt | strip_html strip_newlines | truncate: 160 }}{% else %}{{ site.description }}{% endif %}" />
-  <link rel="stylesheet" href="{{ 'assets/css/main.css' | prepend: site.baseurl }}" />
```
thekaveman, line 9: should we be using `/assets/...` here (e.g. prepend with the `/`) - this is consistent with how we list the assets in e.g. the collections.

_includes/head.html
```diff
@@ -6,7 +6,7 @@
...
+  <link rel="stylesheet" href="{{ 'assets/css/main.css' | relative_url }}" />
   <link rel="canonical" href="{{ page.url | replace:'index.html','' | prepend: site.baseurl | prepend: site.url }}" />
```
thekaveman: This might be cleaner as `<link rel="canonical" href="{{ page.url | absolute_url }}" />`

thekaveman (changes requested):

PR #71, opened 2019-04-29
changes to website images 

stu562:
@thekaveman made the requested revision to changes 
1) edit base url and remove endpoint from url property
2) added realative_url to img src and link tags
3) append link tag to have '/'

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -8,7 +8,7 @@ email: hello@hackforla.org
 description: Hack for LA's website!
 
 baseurl: "/website"
-url: https://hackforla.github.io/website/
+url: https://hackforla.github.io
 
 
 collections:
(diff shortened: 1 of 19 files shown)


abregorivas (approved):

thekaveman:
@stu562 @abregorivas why were some of the `alt` values wrapped in double quotes **and** single quotes? Looks like the ones with spaces e.g. [`alt: "'adopt civic arts'"`](https://github.com/hackforla/website/blob/gh-pages/_projects/adopt-civic-art.md) - but this shouldn't be a problem to just wrap in double quotes right?

stu562:
@thekaveman when implementing just single quotes there was a problem of alt property getting cut off at the first white space
examples:
1) "adopt civic arts" 
2) adopt civic arts
would both results in just an alt = "adopt"

wrapping "'adopt civic arts'"
allows for the whole property to pass without getting cut off

PR #77, opened 2019-05-01
Add open graph markup tags to header

seanholahan:
added open graph markup tags to the header so that sharing HFLA link on social media has img, description etc..


abregorivas (changes requested):
Line 5 should not point to your github repo. It should point to ours.

thekaveman:
> Line 5 should not point to your github repo. It should point to ours.

Even better, we should just use the `absolute_url` filter and never hardcode the URL in the content.

Same with Line 9, the image URL should use a leading `/` and the `absolute_url` filter.

Also, there appears to be a lot of extra whitespace between the `property` and `content` attributes, please clean that up.

abregorivas:
This pull request needs to be re-based from the latest update of gh-pages. Sean, Let me know if you need help with the rebase.

thekaveman:
Closing as superseded by #93

PR #84, opened 2019-05-02
fixed issue #83

abregorivas:
fixes issue where links are being defaulted to site.url if you use the keyword "url" as a variable in the markdown frontmatter.


ExperimentsInHonesty (approved):
This was a necessary change because url is a protected word.

PR #93, opened 2019-05-07
added meta tags for link sharing

seanholahan:
added meta tags for link sharing on social media sites


_includes/head.html
```diff
@@ -2,6 +2,11 @@
   <meta charset="utf-8" />
   <meta name="viewport" content="width=device-width, initial-scale=1, shrink-to-fit=no" />
   <meta http-equiv="X-UA-Compatible" content="ie=edge" />
+  <meta property="og:url" content="{{ page.url | absolute_url }}" />
+  <meta property="og:type" content="website" />
+  <meta property="og:title" content="You can change LA for the better!" />
```
thekaveman: Title should be `Hack for LA`

thekaveman (changes requested):

PR #96, opened 2019-05-08
Accessibility fixes 75

seanholahan:
fixes 3 accessibility issues brought up by "Wave" chrome extension
1. delete empty h3 inside of hacknights location
2 added aria-label to social media icons
3. No label for "send us a message" input


_includes/hack-nights.html
```diff
@@ -11,7 +11,6 @@ <h2>Hack Nights - Every Monday &amp; Tuesday</h2>
             <div class="location">
                 <div class="location-img" style="background-image: url({{ item.image | absolute_url }})"></div>
                 <div class="location-details">
-                    <h3 class="location-header"></h3>
                     <strong>{{ item.title }}</strong> {{ item.event-date }}
```
thekaveman, line 14: I'm actually thinking that the `<h3>` was supposed to wrap `{{ item.title}}` instead of the `<strong>` - can you do that instead?

thekaveman (changes requested):

_includes/hack-nights.html
abregorivas: Actually The original code looked as shown below. If you add the title into the h3 you loose the bold and the title will be a block item vs an inline tag. I think removing the h3 is the way to go.
<div class="location-details"><h3 class="location-header"></h3>
<strong>Westside</strong> Mondays at 6:00 pm<address>
</div>

thekaveman (approved):

Issue #75, opened 2019-04-30
Review and fix accessibility issues using chrome extension "WAVE" tool.


abregorivas:
Steps:
1. add WAVE chrome extension: https://chrome.google.com/webstore/detail/wave-evaluation-tool/jbbplnpkjmmeebjpijfedlgcdilocofh

2. visit gh-page live site https://hackforla.github.io/website/
3. click the extension and review the red flags.
4. Run the same steps for the development site (localhost). Ensure that the chrome extension has the "Allow access to file URLs" enabled.
5. Fix each issue and rerun test until no red flags appear.
6. Commit your changes and make a pull request.

abregorivas:
This feedback comes from Drew de Rosario and each section should be reviewed.
 
Best Practices
- Links to cross-origin destinations are unsafe
	https://developers.google.com/web/tools/lighthouse/audits/noopener
	- when using target=_blank we need to add rel="noopener" to the tag
		ensures that new page runs in a separate process
- Browser errors were logged to the console
	this will fix itself when we get images running

Performance
- Eliminate render blocking resources
	- main.css and babel.js are blocking
		both are located in head.html
		need to add async to tags
		https://developers.google.com/web/tools/lighthouse/audits/blocking-resources
- Minify JavaScript
	- consider minifying babel.js to reduce size
		https://developers.google.com/speed/docs/insights/MinifyResources

Accessibility
- Color Contrast
	<h4 class project-title>  not enough color contrast
 	<a> tags - white background and red foreground
	also buttons
- Form Elements
	do not have associated labels
	https://dequeuniversity.com/rules/axe/3.2/label
		e.g. type="text" on all forms. buttons are self labeling
- Links do not have discernible names
	correcting this will help screen readers
	https://dequeuniversity.com/rules/axe/3.2/link-name
		e.g. use aria-label="to describe what the link is"

thekaveman:
@abregorivas this seems like a duplicate/extension of #64 - can we consolidate notes/tracking of accessibility testing into either this issue or that one (and close the one that isn't needed).

abregorivas:
Closing issue and added Drew's feedback to issue #64

ExperimentsInHonesty:
This issue replaced issue #64 which has now been closed

ExperimentsInHonesty:
- light the way image description needs changing
  - It currently says "no parking sign" . 
  - It should say "Veteran returned from service to civilian life"
- work for la  image description needs changing
  - It currently says "building of city la department"
[...]

PR #108, opened 2019-05-15
#75 removes classname from input and fixes accessibility score on lighthouse

stu562:
Also adds  aria-label property

### DIFF
--- a/_includes/forms/contact-us.html
+++ b/_includes/forms/contact-us.html
@@ -4,13 +4,13 @@
   <p> Are you press and want to get in touch? Are you interested in becoming a sponsor? Want to just
   <a href="http://hackforla-slack.herokuapp.com/" rel="noopener" target="_blank">check us out on Slack</a>? Looking to volunteer? </p>
 <form action="thank-you" method="post" class="contact-form" name="Contact Form">
-  <label for="message" class="sr-only">
+  <label for="message">
       Send us a message
   </label>
   <textarea id="message" name="message" placeholder="Send us a message" required></textarea>
   <div class="form-controls">
     <div class="form-input">
-      <label for="contact-email" class="sr-only">
+      <label for="contact-email">
         Enter your email address
       </label>
       <input
@@ -18,6 +18,7 @@
         type="email"
         name="email"
         placeholder="Enter your email address"
+        aria-label="Enter your email address"
         required
       />
     </div>
--- a/_includes/forms/contact-us.html
+++ b/_includes/forms/contact-us.html
@@ -10,9 +10,6 @@
   <textarea id="message" name="message" placeholder="Send us a message" required></textarea>
   <div class="form-controls">
     <div class="form-input">
-      <label for="contact-email">
-        Enter your email address
-      </label>
       <input
         id="contact-email"
         type="email"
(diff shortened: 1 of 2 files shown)


_includes/forms/contact-us.html
```diff
@@ -4,20 +4,21 @@ <h3> Contact Us</h3>
   <p> Are you press and want to get in touch? Are you interested in becoming a sponsor? Want to just
   <a href="http://hackforla-slack.herokuapp.com/" rel="noopener" target="_blank">check us out on Slack</a>? Looking to volunteer? </p>
 <form action="thank-you" method="post" class="contact-form" name="Contact Form">
-  <label for="message" class="sr-only">
+  <label for="message">
```
abregorivas: We do not want to remove this class. The text should be hidden but readable for screen readers.

_includes/forms/contact-us.html
```diff
@@ -4,20 +4,21 @@ <h3> Contact Us</h3>
...
       Send us a message
   </label>
   <textarea id="message" name="message" placeholder="Send us a message" required></textarea>
   <div class="form-controls">
     <div class="form-input">
-      <label for="contact-email" class="sr-only">
+      <label for="contact-email">
```
abregorivas: Same as above keep class "sr-only"

_includes/forms/hero-signup.html
```diff
@@ -1,7 +1,12 @@
 <form class="hero-signup" name="Quick Signup Form">
   <div class="form-input">
-    <label for="hero-signup" class="sr-only"> Enter your email address</label>
-    <input id="hero-signup" type="email" name="email" placeholder="Enter your email address" required />
+    <label for="hero-signup">Enter your email address</label>
```
abregorivas: keep "sr-only" class

abregorivas (changes requested):
Please add back the "sr-only" classes as they are needed to keep the look of the site the same but still accessible.

stu562:
updated pull request, 
removed label from code. class="sr-only" purpose is to make it not visible. 
the arial label is a text label that is not visible to the screen. Passes audit score and verifying not visible on local enviornment

thekaveman:
hey @stu562, I think @abregorivas has it right - we should prefer `sr-only` hidden labels vs. `aria-label`. I've only done a small bit of reading, but it seems that ARIA properties aren't actually *required* to be supported in Screen Readers.

Here's an interesting look at this exact issue across a range of Screen Reader + Browser combinations: https://www.powermapper.com/tests/screen-readers/labelling/input-text-aria-label/

The conclusion is:

> A more reliable alternative is wrapping the INPUT with a LABEL element.

ExperimentsInHonesty:
I have reassigned the issue to @abregorivas, so that he can either test the code, or discuss in person on next hack night with @stu562 Steven.

abregorivas:
Please remove the label tags for the following 
- contact-us form remove(contact-email, message) label
- hero-signup form remove(hero-signup) label
Add the text that was in labels to an "aria-label" to its corresponding input field.

ajb176:
You must be a member of the HFLA website team in order to create pull requests. Please see our page on how to join us as a member at HFLA: https://www.hackforla.org/getting-started.  If you have been though onboarding, and feel this message was sent in error, please message us in the #hfla-site team Slack channel with the link to this PR.

PR #213, opened 2019-12-05
Projects List Page

StephenVNelson:
So I have fully functioning filtering and the basic design. 
1. There is one issue where the top margin of the footer is making a white space at the bottom of the page. I'm open to suggestions as far as how we handle that because changing the footer CSS would change everything on every page.  (see screenshot below) 
2. When @harishlingam and I talked last week I let him know that currently none of the projects have any information about what programming languages they're using on their projects. So I need that information before I can add the LANGUAGES filter on the prototype.
3. @harishlingam Also said he will get a UX/Design person to look at what I have thus far. I am attaching a screenshot of the whole page so you guys can reference it. (see below)
3. I know we just edited the Not Today project image, but it's still different from all the others because it's a square instead of a rectangle and that's a little annoying, yeah?
4. Do I just submit a pull request now? If you guys accept the pull request will it be added to the main branch of the project? Since we have a few more details that need to be done on the all projects page I'm not sure we want to do a pull request yet? Can we merge my changes into a different branch so that it won't be live quite yet? LMK
FireShot Capture 007 - Hack for LA - localhost


ashlingam:
I think the mockup looks pretty good. As discussed, it will be good to get a once over by a UI lead. We do need to collect languages data from all the projects. I will work on collecting this for you as soon as is possible in a format that will work.

PR #315, opened 2020-02-17
Add project card sort by status

joelparkerhenderson:


cnk:
I included the optimization of the sorting loop into #341 along with moving the project card into its own file.

PR #320, opened 2020-02-19
Sample data from GitHub api

cnk:
Committing a sample data file so Kian can code with actual data while we are fighting with GitHub Actions to generate it dynamically.


cnk:
This is irrelevant because #330 is going to build the real file nightly

PR #357, opened 2020-03-09
Styling updates to project home page

KianBadie:
Some styling updates to the project home page. I don't know why but it says there is a modification in the Record Clearance project file. I copied and pasted the code from the file on the main site into mine but it still says there is a difference.


t-will-gillis:
Dev Team Leads- Dev requests reactivation of their access to 'website-write'

Issue #252, opened 2020-01-07
Update Project Profile: helloGOV


ashlingam:
### Overview
Update the following fields for your project. These updated fields will then be updated to be shown on the website.

### Action Items
Provide in the comments below:
- [x] updated wording if necessary on the project card (see [hackforla.org](https://www.hackforla.org/))
- [x] information on which Tools your team is using for the project (Ex: figma, Balsamiq, photoshop, etc.) . This is different than programming languages
- [x] updates for "Looking For" Section
- [x] any updated links that are not currently displayed on the site (do you have a demo site?)
- [x] Hero image to be displayed on Project Home Page at 1500 x 700 px
- [x] any resources for a Getting Started link (either a link to a wiki or readme)

### Resources/Instructions
See the hero image on the [hackforla.org](https://www.hackforla.org/) homepage

kate-rose:
Update wording of anything currently on the project card (see hackforla.org)

HelloGOV is helping grassroots organizations connect supporters to their state assembly and state senate representatives for call campaigns to advocate on the legislation that matters most to their work. The HelloGOV webapp generates a campaign shortlink that can be used in texts, social posts, and more.

 Update "Tools" Section to display the tools being used for the project (Ex: figma, Balsamiq, photoshop, etc.)

Mongo, Angular, Node, JS

Update "Looking For" Section

Looking for folks to help with implementing end to end testing, user testing and outreach, and a UI redesign. All skill levels, all kinds of contribution welcome.

 Update any links that are not currently displayed on the site (do you have a demo site?) - none
 Banner Image to be displayed on project specific page 16:9 ratio (min width 1200px) - none

 Any resources for a Getting Started link (either a link to a wiki or readme) - https://github.com/helloGov/webapp

ExperimentsInHonesty:
@kate-rose We have to have a banner image, for the project homepages, it's not optional.  Let me know if you need the loan of a UI person to make one, if you don't have someone on your team with that skill set.

Note to website team, here is the url direct to the readme: https://github.com/helloGov/webapp/blob/master/README.md

ashlingam:
@kate-rose  Posted to HelloGOV Slack channel to request update: https://hackforla.slack.com/archives/CDA23KHKP/p1582484825000900

ExperimentsInHonesty:
Use this
![hellogov-hero](https://user-images.githubusercontent.com/37763229/75644569-96023c00-5bf7-11ea-8257-abd2cd20eced.jpg)

[...]

PR #362, opened 2020-03-12
Add leadership section to hellogov.md in projects dir

zempo:
Issue #252


ExperimentsInHonesty:
Hi Solomon, 

I am sending this back to you.   It contains the food oasis file change (see above).  Please fix and I will accept.  Also, you mention the number of the issue it relates to, but don't use the word Fixes so it won't close the issue automatically, like it should.  Don't worry, we will get there.

PR #427, opened 2020-04-12
BUGFIX cards call actual sites "Site" not "Launch Site"

cnk:


cnk:
Kian, should we also add Demo Site to the list of links that show up at the top of the page? If so you could just edit this pull request to add it.

KianBadie:
Fixes #435

PR #455, opened 2020-04-20
update homeuniteus.md

ExperimentsInHonesty:
Updated title and link to repository

### DIFF
--- a/_projects/homeuniteus.md
+++ b/_projects/homeuniteus.md
@@ -1,6 +1,6 @@
 ---
 identification: '228981080'
-title: Home Unite US
+title: Home Unite Us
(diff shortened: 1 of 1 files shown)


yulia-kh (commented):
Hi @ExperimentsInHonesty I noticed that in the description we still refer to the project as Host Home. And I think it's good idea to rename images too  so we won't be confused in the future. Thank you

ExperimentsInHonesty:
@yulia-kh The host home project referred to in the description is about the stakeholders name for their internal process.  It's one of the reasons we decided to rename the hfla project that works on it,  in order to reduce confusion.  I agree that we should rename the images, and I will make another issue to do that, because I think there are two projects that are affected by recent name changes.  Thanks for catching that.

PR #495, opened 2020-05-05
Sort data from GitHub before writing to the file

cnk:
It would be nice to minimize the changes in the json file so that the 'git diff' between versions shows you real changes - not just differences in the order in which the endpoints returned data.

This commit adds a sorting function and then uses it to sort the json objects by project name before writing them to the file

### DIFF
--- a/github-actions/get-project-data.js
+++ b/github-actions/get-project-data.js
@@ -69,6 +69,29 @@ var github = {
     }).catch(function(err) {
       return err.message;
     });
+  },
+  compareValues: function(key, order = 'asc') {
+    return function innerSort(a, b) {
+      if (!a.hasOwnProperty(key) || !b.hasOwnProperty(key)) {
+        // property doesn't exist on either object
+        return 0;
+      }
+
+      const varA = (typeof a[key] === 'string')
+            ? a[key].toUpperCase() : a[key];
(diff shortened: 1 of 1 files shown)


KianBadie:
Ran locally on my machine using node and it seemed like it worked out fine. The data was indeed sorted alphabetically. I also ran the current script that we have locally, and when I compared the 2 results they both had the same amount of lines of data, so I'm assuming there is no data difference. 

Just one question which is: @cnk, I noticed that the location that the data is being written to has changed, and the name of the file has changed. The name of the file I can change myself (because I think changing the name might break the project home layout because it expects the file to be the original name), but I was curious if there was a specific reasoning for changing the location. If so, we would have to roll out the changes to the project home page along with this pull request I think.

KianBadie:
Also, another thing I wanted to note. Not sure if this is the right place. There is a repo called "ocean maps" that appears when I run your script and the orignal script locally. But it is not present on the data on the "github-data.json" on this repo. Unless "ocean maps" adding the hack for la tag within the past 14hrs...

cnk:
Good catch on the filename - I corrected that.
And it does look like there is a new project with the hack-for-la tag: https://github.com/hackforla/ocean-maps

KianBadie:
Looks good to me, I'm going to go ahead and merge this.

Issue #507, opened 2020-05-11
Change sort order for github-date.json


cnk:
We have 2 projects whose repositories are named 'website' so we are still getting unnecessary churn in our github-date.json file. Change the sort order to the 'identification' field which will always be unique.

ExperimentsInHonesty:
@cnk I closed this issue manually because there was no fixes comment on the issue.  Please reopen if its still being worked on.

PR #513, opened 2020-05-14
Sort by project id because we have 2 projects named 'website'

cnk:
We are still getting commits every night because the 2 projects named 'website' and not returned in a deterministic order. Change the sort to be by project id so the sort key is guaranteed to be unique.


KianBadie:
Ran locally. Still the same amount of data and it is indeed sorted by id :+1:

ExperimentsInHonesty:
Fixes issue #507

PR #524, opened 2020-05-17
430 completed projects

### LINKED ISSUE #430: Change logic for project card and project homepage
### Overview
We need to edit what is seen for projects that are completed.  Currently it does not display a project homepage, but we are going to re-enable that, and just edit what does appear.

### Action Items
For projects with a status of completed
- [x] hide getting started link (on project homepage)
- [x] write logic around not displaying getting started on projects with a completed status.
- [x] All project lead links will need to go to their github or linked instead of slack.
- [x] Add the a new field to the project.md file template called "completed-contact"
- [x] Add a project contact email link where open roles used to be. 
   - [x] PM to work with Designers about adding "If you would like to talk to us about this completed project, please reach out to team@hackforla.org"
- [x] Add logic that automatically uses team@hackforla.org as the email if there is not completed-contact email listed.

### Resources/Instructions
[example project with completed status](https://www.hackforla.org/projects/work-for-la)

yulia-kh:
Fixes #430

### DIFF
--- a/_includes/project-card.html
+++ b/_includes/project-card.html
@@ -1,16 +1,26 @@
 {% assign project = include.project %}
 <li class="project-card">
   <div class="project-card-inner">
+    {%- if project.project-homepage == false -%}
+    <div class="project-tmb">
+      <img src="{{ project.image | absolute_url }}" class="project-tmb-img" alt={{project.alt}}/>
+    </div>
+    {%- else  -%}
     <a href='{{ project.url | replace: ".html", "" }}'>
       <div class="project-tmb">
         <img src="{{ project.image | absolute_url }}" class="project-tmb-img" alt={{project.alt}}/>
       </div>
     </a>
+    {%- endif -%}
     <div class="project-body">
       <div class='status-indicator {{ "status-" | append: project.status }}'>
         <h5 class='status-text'>{{ project.status }}</h5>
       </div>
+      {%- if project.project-homepage == false -%}
+      <h4 class="project-title project-title-no-page">{{ project.title }}</h4>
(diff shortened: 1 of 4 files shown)


cnk:
@yulia-kh Generally looks good but can I ask you for a couple of changes? 

1. The email address in the "please reach out to ___" sentence needs to be a `mailto:` link - and for bonus points, would you mind adding a `subject` parameter that says "Question about \<project name here\>" - https://css-tricks.com/snippets/html/mailto-links/
2. I think there may be an occasion for needing project-homepage = false, so I would leave that in project-card.html (and the style in _projects.css)

KianBadie:
Have we confirmed if we want to keep the logic and styling that @cnk mentioned? It also seems like the mailto link is another requirement from the most recent discussion in #524 and from @cnk's comments. 

Overall it looks good, and it meets the requirements listed on the issue (I think, are the first 2 action items in the [description](https://github.com/hackforla/website/issues/430#issue-598558140) of the issue the same thing? Or are they 2 different things?). 

Only 2 notes I had was that linkedin profiles are prioritized over github links. I wanted to double check if that is what we want. The second thing is that the page could potentially display Open Roles and the Completed Contact info if the project card, for some reason, had both info. The project card having both info would be a pretty rare case I think and maybe wouldn't be approved to merge to begin with, but I think it would be safest to make the logic something like
```
if completed and completed-contact
   ...
else if completed
   ...
else if page.looking
   ...
```  
so that it will only display one for sure.

yulia-kh:
@cnk @KianBadie Requested changes implemented

PR #543, opened 2020-05-26
Added Meeting Times to Project Pages Fixes #541

### LINKED ISSUE #541: Add team meetings to project pages
Mock up:
Screenshot from 2020-03-22 10-54-06

Right now, the online meetings list uses the same id/styling rules as the "#open-roles" section. The changes we made to make it look different from the Open Roles section is to make the column-count: 1. In addition, there is a css class/id that adds the styling to the bullet-ed list labeled "#open-roles li:before {}". A quick fix was to remove the "content" property of that class. However, I am guessing a better option would be to either create a new css class for the meetings lists that creates a new styled list of it's own. Or to continue having the meetings list inherit the #open-roles class, and add another class that removes/changes the desired styling. 

My vote goes for creating a new css class that creates a new styled list of it's own.

_Originally posted by @KianBadie in https://github.com/hackforla/website/issues/385#issuecomment-602248280_

ruben1s:
Fixes #541

### DIFF
--- a/_layouts/project.html
+++ b/_layouts/project.html
@@ -61,7 +61,12 @@ layout: default
                     {{ tech }}{% if forloop.last == false %},{% endif %}
                   {% endfor %}
                 </p>
-                {% endif %}            
+                {% endif %}
+                <p class="meetingsHeader"></p>
+                <div id="meetings">
+                    <ul class="meeting-times">
+                    </ul> 
+                </div>           
              </div>
 
             <!--Project Summary and Getting Started Button-->
@@ -200,8 +205,6 @@ layout: default
         languagesSection.remove();
     }
 </script>
-
-
 <!--Script for constructing Team Card-->
 <script>
     let contributors = document.getElementById('contributors-list');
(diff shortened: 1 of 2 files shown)


ruben1s:
Removed console.log on second merge.

_layouts/project.html
```diff
@@ -305,3 +308,197 @@ <h4 style='color:#333'>Resources</h4>
...
+let projectTitle = '{{ page.title }}';
+let meetingsToday = false;
+
+// Assigns local JSON Data to localData variable
+{% assign localData = site.data.project-meetings-from-vrms | jsonify %}
+let localData = {{ localData }};
+
+const API_URL = 'https://www.vrms.io/api/recurringevents';
+
+// Assigns variable to each header by ID
```
cnk: The html doesn't have any of these elements so this code is not needed

cnk (changes requested):
@ruben1s The meeting info looks good on the page but the code contains a lot of stuff that isn't needed (or even used?). Since the output is just a list of the upcoming meetings for this project - not broken out into days of the week, I don't think there is any reason to be creating the day of week lists. Can you simplify this code so it just goes through the JSON data and grabs all the events for the project we are looking for? The data we receive is ordered so I think you can just display it as is - once you have grabbed the relevant entries.

_layouts/project.html
```diff
@@ -305,3 +308,134 @@ <h4 style='color:#333'>Resources</h4>
...
+        scheduleDay(day);
+    }
+    // Adds Meetings: to project page
+    if (meetingsToday === true) {
+                meetingsHeader.insertAdjacentHTML("beforeend", '<strong>Meetings: </strong>');
+        }
+}
+
+// Function removes duplicates in an array
+function removeDuplicates(array) {
```
cnk: Not used

_layouts/project.html
```diff
@@ -305,3 +308,134 @@ <h4 style='color:#333'>Resources</h4>
...
+        let webURL = scheduleData[i].project.hflaWebsiteUrl;
+        let projectName = scheduleData[i].project.name;
+        let description = scheduleData[i].description;
+        let day = new Date(scheduleData[i].date).toString().substring(0,3);
+        let scheduleClass;
+        
+        // Function that adds data to proper header
+        function scheduleDay() {
+            // Assigns different class to data depending if it utilized local JSON file or the VRMS fetch data
+            if (scheduleData == localData) {
```
cnk, line 339: Question: Why are you assigning different css classes to the 2 data types? Is this so you can see if the updates succeeded? or is there some other reason?

_layouts/project.html
```diff
@@ -305,3 +308,134 @@ <h4 style='color:#333'>Resources</h4>
...
+let meetingsToday = false;
+
+// Assigns local JSON Data to localData variable
+{% assign localData = site.data.project-meetings-from-vrms | jsonify %}
+let localData = {{ localData }};
+
+const API_URL = 'https://www.vrms.io/api/recurringevents!';
+
+// GET request 
+let calendarFetch = fetch(API_URL, {method: 'GET'});
```
cnk: Is there a reason to assign this to a variable? It's only used once - and very far away from this assignment which makes the code harder to read.

_layouts/project.html
```diff
@@ -305,3 +308,134 @@ <h4 style='color:#333'>Resources</h4>
         textDiv.style.padding = '35px 15px';
     }
 </script>
+<script>
+// Script for adding meeting times to "meetings" Section
+let meetingsList = document.querySelector('.meeting-times-list');
+let meetingsHeader = document.querySelector('.meetingsHeader');
+let projectTitle = '{{ page.title }}';
+let meetingsToday = false;
```
cnk: NitPick: this variable name is misleading. The meetings are not necessarily today. How about meetingsFound?

_layouts/project.html
```diff
@@ -305,3 +308,134 @@ <h4 style='color:#333'>Resources</h4>
...
+        
+        // Function that adds data to proper header
+        function scheduleDay() {
+            // Assigns different class to data depending if it utilized local JSON file or the VRMS fetch data
+            if (scheduleData == localData) {
+                scheduleClass = "localData";
+            } else {
+                scheduleClass = "updated";
+            }
+            if (projectTitle.toLowerCase().substring(0,4) === projectName.toLowerCase().substring(0, 4)
```
cnk: I would prefer not to work around the differences in project titles between our project cards and the VRMS titles. I would prefer to leave the mismatched projects without meeting times to encourage VRMS to fix the titles - or insist we change our titles. 
But if you want to include this code, it needs a comment to explain the purpose and the implimentation.

_layouts/project.html
ruben1s: I gave it two diff classes so that the js knows what to replace using the class as an identifier.

_layouts/project.html
ruben1s: I removed meetingsToday

_layouts/project.html
ruben1s: Just a good habit. I'll change it.

_layouts/project.html
ruben1s: Removed

_layouts/project.html
ruben1s: I reverted to an === and will wait for VRMS to change their data

_layouts/project.html
ruben1s: Removed

_layouts/project.html
cnk: Ok cool

PR #545, opened 2020-05-27
Update banner styling on project home pages to be more dynamic.

### LINKED ISSUE #510: Prototype new styling for project home page banner images
### Overview
The way the banner images are displayed on project home pages currently have some issues in terms of cropping/zooming in. Prototype new styling rules to better display these images on mobile and desktop (desktop being the trickier of the two).

### Action Items
- [x] Apply mobile styling rules described [here](https://github.com/hackforla/website/issues/255#issuecomment-608226848) by @ddfridley to banner images
- [x] Post screenshots of mobile styling rules
- [x] Experiment with changing pixel dimensions of banner images to something more dynamic with desktop (em, vh, vw, ...)
- [x] Post screenshots of dynamic styling rules

### Resources/Instructions
- Original issue: #255 
- Attempt at summarizing above issue: https://github.com/hackforla/website/issues/255#issuecomment-623010067

KianBadie:
Fixes #510

### DIFF
--- a/_sass/components/_project-page.scss
+++ b/_sass/components/_project-page.scss
@@ -2,17 +2,13 @@
   background-size: cover !important;
   display: flex;
   flex-direction: column;
-  height: 100vh;
   justify-content: flex-end;
-  max-height: 700px;
-  min-height: 100vh;
   padding: 20px;
   position: relative;
   margin-top: 29px;
-
-  @media #{$bp-tablet-up} {
-    min-height: 0;
-  }
+  width: 100vw;
+  height: 46vw;
+  max-height: 74vh;
 
   &::after {
     @include position(absolute, 0 0 0 0);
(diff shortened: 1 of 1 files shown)


_sass/components/_project-page.scss
```diff
@@ -2,17 +2,13 @@
...
-  min-height: 100vh;
   padding: 20px;
   position: relative;
   margin-top: 29px;
-
-  @media #{$bp-tablet-up} {
-    min-height: 0;
-  }
+  width: 100vw;
+  height: 46vw;
```
ye-susan, line 10: For height - was this suppose to be a vh instead vw?
I tested vh instead and the images appear bigger (actual 46% of viewport) while with vw it appears about ~25% of the viewport? Either one looks good, just depends on how big we want the images to appear.

KianBadie: Good question. I kept the height using view width in order to keep the same aspect ratio of the original image. As 100vw/46vw is around a 2.17 ratio (since they both use take different percentages from the same source, the view width). The images we originally asked for were 1500px/700px which is around a 2.14 ratio.  

But since the height of the screen is smaller, I think 100vw/46vh would result in an aspect ratio bigger than the original images (because they are taking different percentages from different sources, so it's still not a 2.14 ratio). I think making both use the vw reduces the amount of cropping the image since it is using the original aspect ratio.

ye-susan: I see, thanks for clarifying!

PR #566, opened 2020-06-07
Project filter

tylerthome:
Marking this as a draft pending discussion with lead developers. Current driver script uses a variety of unstable, not-robust methods of compiling project data into filtering structure, including reads on the github-data JSON and parsing innerHTML on DOM nodes. 

Once a source data representation is specified, the *top-level* script (i.e. inside current-projects.html) should be cleaned up to read only from one well-defined data source. This will also allow some DOM-building logic to be factored out into library code.


cnk:
The way I have my forks set up I am not seeing an easy way to make a pull request without getting other stuff involved. @tylerthome Add this right below your script tag at line 31 and see if this produces the data structure you were getting from pulling apart the html. I tried replacing your code with this but all that happened was the filters disappeared so I am going to pass it back to you for integration.  

Don't worry about the 'looking' content not pulling from our regularized categories. If you can get the dropdowns working with this new data, I'll refactor the 'looking' data to be what we need. I wasn't able to hand parse the lists. So if this code works, we are going to have to rearrange the changes we just made to how we are keeping the "Looking" data to make 2 lists rather than a list of hashes/objects.  

`  var projectDataFromJekyll = [
    {%- for project in site.projects -%}
    {"id": "{{ project.identification }}",
     "title": "{{ project.title }}",
     "data": {"status": "{{ project.status }}",
              "location": "{{ project.location }}"
              {%- if project.looking -%}
              , "looking": {{project.looking | jsonify }}
              {%- endif -%}
              {%- if project.technologies -%}
              , "technologies": {{project.technologies | jsonify }}
              {%- endif -%} }
    }{%- unless forloop.last -%},{% endunless %}
    {% endfor %}
  ];
  console.log(projectDataFromJekyll);
`

cnk:
@tylerthome thanks for creating the filtering mechanism. The code from this PR formed the basis for the pull request that added filters (with styling) to the home page: #619 so I am going to close this one.

PR #572, opened 2020-06-08
Fixes #547

### LINKED ISSUE #547: Javascript error on CivicTechIndex page
### Overview
When you go to https://www.hackforla.org/projects/civic-tech-index you see an error in the JavaScript console. 

`Uncaught Type Error: Cannot read property 'split' of undefined  at civic-tech-index:807`

### Detail about problem
The "Uncaught TypeError"  is happening because there is a line in the list of contributors that is NOT a GitHub user object. Instead it is the string  "Cannot read property 'forEach' of undefined". That is getting inserted because https://api.github.com/repos/civictechindex/database/contributors does not return any data. 

### Action Items 
We need to update the data retrieval script to deal with projects that do not yet have contributions.

### Resources
There are two errors on the page.  Ignore the one that says "A cookie associated...", because it happens on all project pages.

ruben1s:
Cleared console of errors on project pages stemming from invalid github contributors data.

### DIFF
--- a/_layouts/project.html
+++ b/_layouts/project.html
@@ -215,26 +215,40 @@ layout: default
 <!--Script for constructing Team Card-->
 <script>
     let contributors = document.getElementById('contributors-list');
+    // Empty array for valid contributors
+    let contributorsArray = [];
+
     if(project != null){
-        for(contributor of project.contributors.data){
-            let contributorDiv = document.createElement('div');
-            contributorDiv.classList.add('contributor-div');
+        for (contributor of project.contributors.data) {
+            // Checking for valid data
+            if (contributor.github_url && contributor.github_url != "https://github.com/apps/dependabot") {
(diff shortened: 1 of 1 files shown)


KianBadie:
Nice, so this does fix the error. And through the fix, it is showing the last contributor that was being skipped because the error was thrown before it could get to the last one. 

The action items in the issue ask for the data retrieval script to be changed (get-project-data.js), and this modifies the script that handles that data after it was fetched. Still, it is a fix to the error with positive results (1 more contributor showing). @cnk would we want to merge this and fix the data retrieval script after?

cnk:
I had been thinking of fixing the data script but this fix does get me what I want (no errors in the JS console when looking at the site) and I like that @ruben1s filtered out Dependabot. I wonder if we should consider filtering out HackForLABot too. But that's for another issue.

PR #621, opened 2020-07-21
added Overview pdf file functionality

jbubar:
changed overview picture to match project img and added overview to the getting started drop down


KianBadie:
Looks good as well!

PR #624, opened 2020-07-21
Change README.md to clarify instructions for accidentally cloning hac…

drubgrubby:
In the instructions for what to do if you accidentally clone the hackforla/website.git rather then cloning from the user's online repository it was a little confusing if you should do the next instruction, or the next two instructions.  I added some language to (hopefully) make that clearer. 

Also, at Bonnie's suggestion, changed the naming convention when creating a new branch so that it doesn't have to start with the issue number.

### DIFF
--- a/README.md
+++ b/README.md
@@ -100,13 +100,15 @@ Verify which URL your `origin` remote is pointing to:
 git remote show origin
 ```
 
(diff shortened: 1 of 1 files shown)


cnk:
@drubgrubby Thanks for the update and congrats on your first contribution!

Issue #459, opened 2020-04-21
Tell us the Technologies (not languages) your project using


ExperimentsInHonesty:
### Overview
Volunteers want to know what frameworks projects use, so that they can better choose which project to join.

### Action Items
Write a comment in this issue with:
1. the name of your project
2. the frameworks you use (e.g., React-JS, Vue, Angular, Ember, Backbone, Express, Django, Rails, etc.)

ddfridley:
1. Undebate
2. Mongo, Express, React, Node (MERN)

theswerd:
1. New Schools Today
2. ReactJS, Flutter

eselkin:
1. Engage
2. React, Django, SQL (PostgreSQL)

c-fuchs:
1. HomeUnite.Us
2. React, NoSQL

ExperimentsInHonesty:
I am going to message each of the projects directly throughout the day.

311-data slack message: https://hackforla.slack.com/archives/CJY4R382X/p1587920983101900
FOLA slack message: https://hackforla.slack.com/archives/CMER3R1RD/p1587922553171400
HelloGov slack message: https://hackforla.slack.com/archives/CDA23KHKP/p1587922621006600
Metro-Ontime sent cam a private message since project is completed.
Write for All slack message: https://hackforla.slack.com/archives/CTJN4HPRA/p1587925379142200
TDM calculator slack message: https://hackforla.slack.com/archives/CKY65G266/p1587925315140100
Lucky Parking slack message: https://hackforla.slack.com/archives/CPFDMSV6V/p1587915685024500
Public Tree map slack message: https://hackforla.slack.com/archives/C9UU5KKAA/p1587925293029600
VRMS slack message: https://hackforla.slack.com/archives/CRGH5HM0Q/p1587925343005000

ExperimentsInHonesty:
1. Hack for LA website
2. gh-pages, jekyll

@cnk question for you.  Why does it not show liquid as a language?

breeze-094:
1. Lucky Parking
2. React-JS, PostgreSQL

cnk:
> @cnk question for you. Why does it not show liquid as a language?

Because we do not use a .liquid file extension for our template files.

johnr54321:
For 311:
Javascript: React / Redux
Python: Flask (Sanic) / Pandas
CSS: Bulma
DB: PostgreSql

wesrowe:
[...]

PR #634, opened 2020-07-26
added technologies used for tdm-calculator

jbubar:
#459 
update the project card for tdm-calculator to include technologies used on the project.

### DIFF
--- a/_projects/tdm-calculator.md
+++ b/_projects/tdm-calculator.md
@@ -58,7 +58,16 @@ links:
 looking:
   - category: UI/UX
     skill: UI Designer
-location: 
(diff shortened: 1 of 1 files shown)


KianBadie:
@ExperimentsInHonesty @jbubar I removed "HTML" from the technologies list for consistency with keeping the technologies section to tech so that it wouldn't add a language to the "Technologies" filter. [This](https://github.com/hackforla/website/pull/636#issuecomment-664705466) is also related. Let me know what you guys think

jbubar:
Thanks Kian! I think that is a good move

PR #636, opened 2020-07-26
Added technologies used for food oasis

jbubar:
#459
update the project card for food oasis to include technologies used on the project.

### DIFF
--- a/_projects/food-oasis.md
+++ b/_projects/food-oasis.md
@@ -66,7 +66,6 @@ technologies:
   - Docker
   - Mapobx
   - Material-UI
-  - HTML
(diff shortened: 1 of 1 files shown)


KianBadie:
I noticed that this project and other projects listed HTML as a technology and @jbubar did not include it (which makes sense because this is not the language section). Should we create a new issue for adding languages to these project cards that were mistakenly listed for this issue @ExperimentsInHonesty? Let me know and I can even make it myself. I went ahead and accepted the pull request anyways.

Also, side note: All this new tech is making the filter options really long. Maybe we can make it scrollable? Or just leave it.

PR #639, opened 2020-07-26
adds technologies used by record clearance project

jbubar:
#459
update the project record clearance project to include technologies used on the project.


KianBadie:
@jbubar Watch out for the indentation on the fields for the project cards. The technologies field was indented by one and the card could not load. For future reference :)

Issue #542, opened 2020-05-25
Create Organizational Dashboard page


ashlingam:
### Overview
Create Organizational Dashboard page to house analytics collected from surveys. Initially, the dashboard will house data from the remote onboarding, new member, and project leader surveys, but eventually, expand to include the visualized results from the other surveys in our repertoire.

Instructions to remove the vertical scrollbar have been included below. Please add/remove vertical scrollbar depending on UX considerations.

### Action Items
- [x] Design and populate Google Data Studio dashboard
   - [x] Create dashboard in Google Data Studio
   - [x] Get feedback on dashboard
   - [x] Create dashboard page and insert GDS dashboard via iframe 
   - [x] Revise dashboard
   - [x] Refine page UI in Figma
   - [x] Sign off
- [x] Review feasibility with Kian
- [x] Review with Design team which of the two designs work best with our current design schema and if we are going to leave off the white band at bottom?
- [x] Development
   - [x] Implement page design with Iframe from Figma mockups.  Note Figma mock up missing while band at bottom.  Decision will have already been made to include or not include it.
   - [ ] Show version with scroll bar to Bonnie
      - [ ] decide to remove scroll bar (creating a lot of white space), or leave.
   - [x] complete page.

### Resources/Instructions

Figma https://www.figma.com/file/0RRPy1Ph7HafI3qOITg0Mr/Hack-for-LA-Website?node-id=811%3A1003

#### Desktop Embed:

<iframe width="1200" height="1800" src="https://datastudio.google.com/embed/reporting/16J9XI19pQejSDNFvjl_V3ZezJ502rM_a/page/qx6RB" frameborder="0" style="border:0" allowfullscreen></iframe>
  
#### Mobile Embed:

<iframe width="350" height="2000" src="https://datastudio.google.com/embed/reporting/1xNlu1xcL5y6jh95lX3xr0zb2FCqD92Gm/page/gIfSB" frameborder="0" style="border:0" allowfullscreen></iframe>

#### Remove Vertical Scrollbar Instructions:

https://www.sitepoint.com/community/t/removing-vertical-scrollbar-in-iframe/22495

#### Read More: 

https://support.google.com/datastudio/answer/7450249?hl=en
[How to export one chart at a time from google sheets (provides html)](https://support.google.com/docs/answer/1047436?co=GENIE.Platform%3DDesktop&hl=en)

Something you probably wont need, but good to have
[...]

PR #640, opened 2020-07-26
542 Organizational Dashboard

efrenmarin45:
Fixes issue #542


ExperimentsInHonesty (commented):
I recommend a squash before you do pr but other than that it was fine.

PR #652, opened 2020-07-28
adds technologies used for public tree map

jbubar:
#459 
adds technologies used for public tree map


rblaz001:
To stay consistent with the other filters you should change **Node** to **Node.js**

PR #662, opened 2020-07-31
Fix alt text for .md files that were incorrectly formatted - 625

### LINKED ISSUE #625: Fix "alt" text values in project.md files
### Overview
For a couple of project.md files, the text value of the "alt" text produces errors in how the alt text is actually produced on the website (see image bellow). Verify that alt text is correct on the site for each project and if not, fix the "alt" text in the project.md cards so that it is correct. 

Another route could be to change the way Jekyll incorporates the alt text in the _includes/project-card.html. Maybe putting it in quotes would help. I think this would still require editing the Project.md files though to fix other undesirable outputs though. Just double check the produced alt text if you do it this way too.

### Action Items
- [x] Change alt text values on project cards relative to this issue
- [ ] (Not required, but may be useful) Change the way _include/project-card.html incorporates alt text
- [x] Verify all project cards produce correct alt text on the website.

### Resources/Instructions
- To verify the alt text, you can inspect the element and see in the DOM a attribute named "alt" for that image. Here is an incorrect example (Civic Tech Index):
Screenshot from 2020-07-20 21-45-59
Here is a picture of a correct example (Engage):
Screenshot from 2020-07-21 22-20-59
- I know 311 Data, Civic Tech Index, and Civic Opportunity Project produce incorrect alt text for sure, so that can be a starting point. Refer to project.md files like Engage to see correct implementation.

drubgrubby:
Fixes #625 

- Corrected image alt attribute so that all cards show alt text in correct format. 
- I noticed that 311 and tdm-calculator didn't have an alt-hero attribute, so I added them. 

Note: I noticed that the main pages for each card, where the hero-image is displayed, do not display the alt-hero text.  I ran grep on the directory and didn't see the text alt-hero anywhere other than on the .md files, so perhaps this is known or intentional.  Just mentioning it in case...

### DIFF
--- a/_projects/100-automations.md
+++ b/_projects/100-automations.md
@@ -3,9 +3,9 @@ identification: '67671778'
 title: 100 Automations
 description: Hack for LA seeks to reduce repetitive work in our open source projects and for the open source community.  This project will be both a showcase for the automations and/or microservices that we develop, but also a convenient home for those automations, so that they can be found, forked, and contributed to easily.
 image: /assets/images/projects/100automations.png
-alt: "Open Source logo with hackforla logo on the left edge of it."
+alt: "'Open Source logo with hackforla logo on the left edge of it.'"
 image-hero: /assets/images/projects/100automations-hero.png
-alt-hero: "Green gears with a red center on one"
+alt-hero: "'Green gears with a red center on one'"
 leadership:
   - name: Niven Prasad
     role: Product Manager
--- a/_projects/311-data.md
+++ b/_projects/311-data.md
@@ -3,8 +3,9 @@ identification: '190321758'
 title: 311 Data
 description: The 311 Data project seeks to empower local Neighborhood Councils to improve the ideation and analysis of their initiatives using the wealth of publicly available 311 data.
 image: /assets/images/projects/311.jpg
-alt: "image of building with windows that look like they spell out 311"
+alt: "'Image of building with windows that look like they spell out 311.'"
 image-hero: /assets/images/projects/311data-hero.png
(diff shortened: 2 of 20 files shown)


KianBadie:
I reviewed the code, and it does indeed fix the problem! Unfortunately, after some thought and discussion, I think there is a better solution that is minute in difference but I think better in practice.

KianBadie:
I reviewed the code, and it does indeed fix the problem! Nice job, and thank you. Unfortunately, after some thought and discussion, I think there is a better solution that is minute in difference but I think better in practice.

Right now, jekyll is inserting the alt text like this: ```alt={{project.alt}}``` which turns into ```alt=Some alt text here``` (this example is assuming the alt field in the project.md card is written as ```alt: Some alt text here```), which, as you solution fixes, is wrong since html needs quotes around attribute names. 

Putting double quotes like this ```alt="{project.alt}}"``` allows the alt text to be wrapped in double, single, or no quotes (even though I think there should be a consistent standard this stuff is entered in). So the change would involve surrounding the jekyll statement in quotes like the above example and changing the all the alt text in the project cards to be surrounded in either single, double, or no quotes (just as long as they are consistent).

Sorry to waste your time, I should've just stuck with that when I made and described the issue. Would you mind implementing this solution? And maybe on a new branch or merging the commit messages (not in the way I described a couple days ago haha, in a safe and best practices way) so that the commit log on the site can be kept as minimal as possible? Please let me know your thoughts.

Thank you again!

drubgrubby:
No problem.  Always happy to make a fix better.

PR #673, opened 2020-08-06
651 getting started bug

### LINKED ISSUE #651: Getting Started Page Accordion and Whitespace
### Overview
The bottom of the 'Getting Started' page has a white stripe above the footer that shouldn't be there. The accordion resizes to a bigger width once the user clicks on the chevron when it should remain the same. 
### Action Items
Review previous commits and find out what caused the changes on the 'Getting Started' page.

efrenmarin45:
Fixes #651


KianBadie:
It looks like there are some fragments of code that were left behind by a merge in the _footer and 2FA scss files. I think it would be a good idea to remove those. 

The margins you added does indeed move the Getting Started header section to a spot that's easier on the eyes, so nice job. I want to point out the new margins reveal a grey bar above the getting started header section, and it also looks like a slightly different size grey bar is bellow the getting started header section. It looks like this is from the background color of the #app id css rules. I think it would be a good idea to cover those up some how. Maybe by some how extending the header section to cover those gap areas (if they are even gap areas, I'm not entirely sure what's making them do that)?

Would you have the time to add these changes? And also last thing I would recommend is squashing the commit messages into one message just so we can keep the commit logs a bit cleaner. I think usually it's not the biggest deal if it a couple extra commit messages, but I think the commits like "Merge pull request #1 from hackforla/gh-pages" qualifies the commit messages to be merged into one. Feel free to message me on slack if you make a new pull request and I'll hop on it to review asap.

ruben1s (approved):

efrenmarin45:
@KianBadie Fixed grey bar issue. Appears it was just a simple styling issue with padding and margins. Uploaded image to issue for reference. I removed the commented out code on the Footer and 2FA Sass files to clean up the code. Unable to squash all commits into one on my end. Will be reaching out to you on Slack to get more assistance on this for future commits.

PR #675, opened 2020-08-08
Debug proj filters2

### LINKED ISSUE #645: Fix project filter display on mobile Safari
### Overview
The project filters do not display correct on the mobile version of Safari, so we'll have to create a workaround for mobile phones for displaying the project filters

### Action Items

- [x]  Hide filters on mobile Safari (issue #646 )
- [ ]  Render project filters on touchscreens without using :hover (which causes some inconsistency in performance on some browsers like Safari)

ye-susan:
fixes #645 
This debug includes the following changes
- detect whether device is non/touchscreen
- created a "click" event listener for touchscreens to display dropdowns

Please note, this solution has potential to still be buggy in that there may be double clicks -- but I still want to test this solution out first since it's less verbose than my other fix. 
If there are bugs when we test on different devices, I'll re-do these changes with the more verbose solution.


leonelRos:
@ye-susan The filters work great on Chrome and Safari for both web and mobile.

ye-susan:
@leonelRos Awesome, thank you for looking it over! Do you think it's good enough to merge? If anything, I can go back and fix any additional bugs after testing Safari on an iPhone

leonelRos:
Yes, I can go ahead and merge them after Bonnie approves it.

ExperimentsInHonesty:
@ye-susan @leonelRos this sounds good to me.  Go ahead and push it, so that we can test during meeting across everyones devices.

PR #681, opened 2020-08-13
Get rid of hack nights in nav

theswerd:
Currently the hack nights section is not being shown as hack nights are not active. Therefore when someone clicks "Hack Nights" on the nav they aren't taken anywhere, and it seems like a bug.

For the time being "Hack Nights" should be commented out, until we bring hack nights back

### DIFF
--- a/_data/navigation/main.yml
+++ b/_data/navigation/main.yml
@@ -1,7 +1,7 @@
-- name: Hack Nights
-  link: "/#hack-nights"
-  list_class_name:
-  icon: /svg/icon-hacknights.svg
+# - name: Hack Nights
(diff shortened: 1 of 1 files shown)


ExperimentsInHonesty:
Thank you for this.  We will be accepting this after we make another change.

KianBadie:
@ExperimentsInHonesty Are we able to close this since `Hack Nights` is not in the nav anymore?

KianBadie:
Through discussion with @alexandrastubbs we are closing this pull request as Hack Nights is no longer linked in the nav bar (as this pull request was trying to do). Apologies, @theswerd for lagging on this pull request

PR #685, opened 2020-08-17
Credits page

### LINKED ISSUE #571: Create Credits page
### Overview
Create credits page for website. This page will house attributions for all images, graphics, and photographs used.

### Action Items
- [x] Content strategy
   - [x] Research attribution requirements for different image sources on the web. Identify pricing and attribution placement requirements. 
   - [ ] Purchase subscription (if appropriate) to image source web site
- [x] UI/UX
  - [x] Create mockups for desktop and mobile
  - [x] Revise mockups based on feedback
  - [x] Hand off to dev
- [ ] Development
  - [ ] Implment new Credits page
  - [ ] Insert attributions for all images on Credits page

### Resources/Instructions
https://www.figma.com/file/0RRPy1Ph7HafI3qOITg0Mr/Hack-for-LA-Website?node-id=1074%3A0

ghost:
Fixes: #571

Page is located at credits.html, which displays content from credit-items.html.

Formatting may not be perfectly aligned with [the Figma file](https://www.figma.com/file/0RRPy1Ph7HafI3qOITg0Mr/Hack-for-LA-Website?node-id=1074%3A0), but it looks appropriate and follows standards of the Hack for LA site. 

Screenshots:

image

image

image

Does not fully contain the content because the Figma file seemed to be out of date and did not contain all of the most recently updated credits.


KianBadie:
@ethanh256  Looks good from the screenshots. Nice job, Ethan. Before I clone this branch locally to review, can you please clarify for me the change in the .gitignore file to .gitignore and .gitignore 2?

credits.html
```diff
@@ -0,0 +1,3 @@
+---
+---
+    {%- include credit-items.html -%}
```
cnk: Is there a reason this is a separate file since all it does is include credit-items.html?

ghost: Good point. I can fix this

ghost:
@cnk moved `credit-items.html`'s content into `credits.html` and deleted `credit-items.html`.

@KianBadie Not sure why I made a `.gitignore 2` so I deleted it.

_sass/components/credit-items.scss
```diff
@@ -0,0 +1,267 @@
...
+  font-size: 26px;
+  display: flex;
+  max-width: 895px;
+  text-align: center;
+  padding: 30px;
+  margin: 12px auto;
+  font-family: "Roboto" sans-serif;
+}
+
+.getting-started-page {
```
cnk: Copy and paste from getting-started-page.scss Is this still the the same is the similarly named div? When I compare them, I see one change - but does that change affect the credits page? the getting started page? both? 

If this were a generalized component, it might be OK to reuse - but then it should probably be moved to a file named for the general kind of component this is. This is important to signal to other developers that this is a reusable (and reused) piece of CSS which you should definitely use - but be very careful when modifying.

credits.html
```diff
@@ -0,0 +1,72 @@
...
+                  {%- for item in site.credits -%}
+                        <li>
+                          <!-- originally class="project-card-inner" -->
+                        <div class="credits-inner">
+                          <div class="icon-left">
+                            <img src={{ item.image-url | absolute_url }} class="step-img" alt={{item.alt}}/>
+                          </div>
+                          <div class="icon-info">
+                            {%- if item.title -%}
+                             <div class="project-partner">
```
cnk: Again, these css classes scream cut and paste. This is going to cause problems later because people changing the styling for ".project-parner" are not going to check that the credits page still looks OK after their changes.

cnk (changes requested):
The setup you created - a collection for credit items and then a page to display them - is well done. Unfortunately there are too many issues with the details of your html and scss files for me to be able to merge this at the moment. 

Both files were both clearly created by copy, paste, and modify. That's a fine technique but you need to finish making changes to separate these copied files from their models. In the html file, please give the css classes names that make sense in the context of the credits item - and then propagate those new names into the naming in the scss file. 

If you want to reuse the styles without modification, you should give the style a fairly generic name, and then change both your page and the other pages that use the style to use the new name.

If you want to reuse a style with modification, you can use SASS's mixins to move the shared rules into a mixin - and then include the mixin in both specific css classes where you need it. Docs: https://sass-lang.com/documentation/at-rules/mixin For example the orignal getting-started-page.scss has a class .getting-started-page. In your version of it in credit-items.scss, you have nearly all the same rules and added a margin-top rule. So create a mixin with the share styles and then include that mixin in the css class named for the page:

```
@mixin blue-section {
  background: #030c2c;
  min-height: 100vh;
  display: flex;
  flex-direction: column;
  justify-content: center;
  align-items: center;
  font-family: "Roboto"sans-serif;
  font-style: normal;
  font-weight: normal;
}

.getting-started-page {
  @include blue-section;
}

.credits-page {
  @include blue-section;
  margin-top: 29px;
}
```
Once those issues are taken care of, please resubmit this issue. There are problems with the visuals (this page is not responsive), but I would like to address those after we have some of the engineering aspects taken care of.

ExperimentsInHonesty:
@ethanh256 @cnk on this issue, Cynthia is right this is totally because of the cut and paste that we did.  So my advice is to make a new class that copies the class you are currently using.  But put it in the main css document and make it more generic, so that we can use it in multiple files and we can revise the file we took that from that uses project-parnter later.  @cnk do you agree?

KianBadie:
@ethanh256 One thing I would like to add right now is that I noticed this is a branch on the hackforla/website repo. Usually, work is done on a forked repo. Was this intentional to be on the main repo? I think working on a forked repo reduces the chances of unintentional changes (accidentally pushing to the wrong branch is a rare error I can think of, but on a forked repo has minor consequences. On the main repo however, it can result in the changes going live).

I also wanted to add that I like the idea of making each credit an item in a collection. It's nice!

KianBadie:
@cnk @ExperimentsInHonesty given that component-izing the site is currently being worked on, what would be the best way to go about fixing the classes. The 2 options it seems are creating specific classes in the credit-items.scss file (even a good amount of classes are similar or identical to existing ones), or to make the generic mixins that will be extended in the scss files. I feel like because component-izing the website is currently happening, that creating specific classes in the credit-items.scss file would be easier and quicker, as making the general mixins would create a new standard that seems like it would be overwritten. Thoughts?

**Edit**: @ethanh256 I just finished doing my first round of actual review of the code. I just want to +1 with what @cnk mentioned about the classes. Reusing the same exact classes can have unintended results in the future if they are not generic. I think the mindset of reuse is the mindset to have though. One thing I saw happening with the classes is that a certain class `.project-list` is defined in credit-item.scss as well as _projects.scss. Since they are defined in both places, the current projects section is getting css rules from **both** defined classes. So (for just a silly example) if `.project-list` in credit-items.scss has `padding: 10000px`, that would also unintentional show up on the current projects section on the home page. So I definitely think some class reworking is necessary. 

Another thing I wanted to add is that `.project-partner`, `.project-needs`, and `.project-location` do not seem to be defined classes anywhere. If we remove the class statement that uses them in `credits.html`, then maybe we could delete the `<div>` tags that were using those classes and maybe replace it with a p tag or a span tag? I feel like that would improve readability instead of having to read through about 4 divs. I feel like my eyes can parse `<p>` tags quicker than `<div class="...">`. And _maybe_ it can be single line statements since there is less going on. That would improve readability even more!

What you have here is solid man, keep it up~

ghost:
This pull request is no longer valid and has been replaced by #721. 

**_Closed._**

PR #696, opened 2020-08-22
Add gpl-2.0 license

cnk:


ExperimentsInHonesty (commented):
We should definitely have a licence 
I think with the license type, we have to to scrape everyone who ever contributed to the project and @mention them, that we are going to slap a license on thing and do they object.  And if they do, to let us know which code they contributed to that we should remove.

I’ll create an issue and ask Raudel and Kian to help with that, and once we reach X comment period without any objections we can commit your PR

KianBadie:
@ExperimentsInHonesty was this blocked by #705? If so, is this good to go, or are we still waiting for something? I saw that in #699 it says that the license will be changed on September 27, 2020, which I'm guessing is a typo. I know you just made the comment tagging all the contributors a couple days ago, so was that supposed to mean October 27, 2020?

PR #711, opened 2020-08-29
Issue 651 Getting Started

efrenmarin45:
Fixes Issues #651


KianBadie (changes requested):
Thank you for working on the issue. The fix itself is almost there, I just have one tiny nit pick I will share later. Before I address that, I wanted to address some other things:
- 5 files other than `getting-started-page.scss` got changed, and it looks like these changes were added by mistake. My guess is that the intended fix/changes (which are in `getting-started-page.scss`) were not created on a clean branch. It seems that the work was done on a branch that was not in sync with the current state of the main gh-pages branch. I'm not sure what your preferred workflow usually is, but I would highly recommend that every time you work on a new issue, to create a new branch from a completely up to date/in sync version of the gh-pages branch and to pull in updates to upstream gh-pages before making a pull request. That way, the changes you are creating are branching off in a place where the differences in code will be the least, and pulling in missed updates that went on while you were branched off doing work. With this pull request, the only changes coming in should be in `getting-started-page.scss`. The fastest way to fix this for this pull request is to make your gh-pages branch completely in sync/up to date with the main repo gh-pages, create a new branch from that, add the getting-started mobile fixes to that branch, and then make a pr for that branch. I think doing this will also remove a lot of the commit messages that are showing up above this comment. I'm not sure if those messages would have been squashed with a merge, but I think this whole bullet point still stands.
- For the fix itself, it does get rid of the grey bar :+1: However, the padding rules on it are uneven as it is applying a padding of 15px to the right side, but 0px of padding to the left side. In addition, it is changing the padding on the bottom from 30px to 15px. Remember, if a padding or margin property has 4 arguments it applies those arguments clockwise like this `padding: top right bottom left`. If it has 2 arguments, it applies like this `padding top/bottom left/right`. In regards to the bottom padding changing, I think the safest and quickest way to change only what is needed. So I think it would be good to keep the padding on the bottom at 30px and the padding on the left/right to 15px. On that same note, I think it would also be a good idea to keep the "Getting Started" text in banner to be in the same spot it would have been with the grey bar that was there before (which seems to happen if padding top is calculated to ~87px). Even though it's small, I personally think this mindset is good because it separates stylistic changes with implementation. Like trying to keep the look the same, but change how it is implemented. 
- Lastly, it doesn't look like margin-top rule in `getting-started-page.scss` is changing anything except showing the slightest amount of the top grey bar. I think it would be fine to delete it. If we can delete unnecessary code, it increases readability. 

This is more of a side note since the fix to the first bullet point should make this one irrelevant, but I wanted to add it because I think it will help on future git work. In the other 5 files that are bringing in changes, it looks like there are fragments from resolving merge conflicts (I will comment on some of them in the Files changed tab so you can see what I am talking about). If you use VSCode and select one of the "Accept Incoming/Current Changes" button on a merge conflict, it should get rid of the stuff like  "<<<<<<< HEAD". I think those are just there during a merge conflict to show which code already exists, and which code is coming in that is causing the conflict. But again, if the fix in the first bullet point is implemented, it would only make this bullet point something to keep in mind for the future. Nothing is needed in this pull request in regards to this paragraph is the first bullet point is fixed.

Thank you again, Efren. Let me know if you need anything

_includes/current-projects.html
```diff
@@ -1,5 +1,6 @@
 <a class="anchor" id="projects"></a>
 <section class="content-section projects">
+<<<<<<< HEAD
```
KianBadie, line 3: Example of a fragment from resolving merge conflict

_includes/current-projects.html
```diff
@@ -52,18 +78,36 @@ <h2 class="project-header">Current Projects</h2>
         touch = false;
     }
 
+=======
```
KianBadie, line 81: Another example of a fragment from resolving a merge conflict

efrenmarin45:
Opened a new branch with proposed changes. Closing this PR for clarity.

PR #712, opened 2020-08-30
651 getting started fixes

### LINKED ISSUE #651: Getting Started Page Accordion and Whitespace

efrenmarin45:
Fixes #651


KianBadie:
@efrenmarin45 thank you for all the hard work you've done with this one. Your fix for the nav bars are good.

Good initiative on the chevron fix :+1:  However, I'm afraid I won't be able to merge the pull request with them because I was able to still create the misaligned chevrons by adding more text (I put screenshots below). So I don't think the underlying issue is being solved. I would also have to ask Bonnie if (stylistically) the decreased font size and the padding added to the getting started cards would be cool. If you want to merge this and focus more on getting the issue done I can quickly revert this back during the merge and this issue can be raised up or you can work on it in a new pr. But if you still want to work at it in this pr, that is fine with me too and I can review afterwards.

**Edit:** Also, this is really minor but completely my bad and I'm picky: but the padding top needs slightly different values for each mobile device in order for it to have the same position as before. It looks like 80px is a good inbetween. Like I said, it is super picky and I told you the wrong dimensions, so that's completely on me. I can quickly change that when I merge it. I'll also probably remove the margin-top line since it dons't seem to be changing too much.

**Current Fix**
Screenshot from 2020-08-30 17-47-03
**Adding more text**
Screenshot from 2020-08-30 17-47-14

KianBadie:
@efrenmarin45: Directions to get approved: Revert the chevron fixes back to the original code so that the only fixes present are for the gray bar of the getting started card.

efrenmarin45:
Corrected PR to match @KianBadie requests. Reverted chevron changes, added to issues.

KianBadie:
Thank you @efrenmarin45, looks good. Going to merge

PR #714, opened 2020-09-03
Lack of preloaded data made meeting times heading disappear

cnk:
This fixes the issue opened on the VRMS issue queue https://github.com/hackforla/VRMS/issues/232#issuecomment-685863172


ruben1s:
Does this need to be fixed on the vrms team?

cnk:
The data we are getting from the VRMS team in the AJAX request is correct - and the data we retrieve nightly is correct. However, there was a name mismatch between the file the meeting pages were using and the file we were storing each night via GitHub Actions. This PR fixes the templates to use the file that gets updated nightly.

I did not figure out why not having any meetings on the first load makes the meetings header not appear when the AJAX data brings in new meetings. We might want to open a separate issue to address that.

ruben1s (commented):
Looks good.

NivenPrasad:
Thank you @cnk and @ruben1s for working on this. Looks like the meeting times header now appears for our project (100Automations).

ruben1s:
No problem 👍

Issue #577, opened 2020-06-14
Componentize the website


ExperimentsInHonesty:
### Overview
We need to make the website more modular easier to design and develop new content.   We are going to start with reusable components

### Action Items
- [x] UI/UX
   - [x] Create a new page in Figma that is all the component examples that make up a webpage on our site.
      - [x] Headers: different layout types
      - [x] etc.
  - [x] Review with Design team to determine ideal choices
     - [x] Create style guide.
- [x] Development
   - [x] scss files for web components (e.g., headers, white box in body, etc.
   - [x] see notes from https://github.com/hackforla/website/pull/720 regarding removing mixins vs bootstrap (decision has not been made but can be discussed).

### Resources/Instructions
[Main figma file](https://www.figma.com/file/0RRPy1Ph7HafI3qOITg0Mr/Hack-for-LA-Website?node-id=352%3A205)

daniellen00:
Currently working on this- will have a draft ready for comments/edits

daniellen00:
**Progress:** Completing this- It's taking longer than expected because the website changed compared to the Figma pages, so I have been "inspecting Element" on the site to find new measurements.
**Blocks:** Limited time
**Availability:** Thursday, Friday, Saturday, Sunday
**ETA:** Trying to finish by tomorrow 6/6

daniellen00:
First draft: https://www.figma.com/file/0RRPy1Ph7HafI3qOITg0Mr/Hack-for-LA-Website?node-id=1695%3A13302

Still need to adjust several parts, but getting there!

ExperimentsInHonesty:
@ye-susan will coordinate with @daniellen00 to get consistency - so that page title and other heading names are changed to be what is reflected in the existing typograph scss https://github.com/hackforla/website/blob/gh-pages/_sass/elements/_typography.scss

daniellen00:
Utilized Sass Github folders for consistency between website and Figma per Susan's suggestion, and consulted with her on questions.

Also discussed [Storybook option](https://swizec.com/blog/build-a-new-design-system-in-a-couple-afternoons/swizec/9444) (via Cynthia) for a 'living style guide' with Susan. Will consult with the developers about this more today.

daniellen00:
Determined with Susan that Storybook and other automatic component generating programs like Styleguidist/Docz won't work for the site after all (mostly because they're React-focused).

daniellen00:
[...]

PR #720, opened 2020-09-06
Remove Bourbon SASS framework

cnk:
We were using relatively little of Bourbon. Removed it and added the few mixins we were using to _sass/core/_mixins.scss.

When we start using Bootstrap, we will probably want to delete those in favor of including the Bootstrap versions.

This is a step on the way to componentizing the website: issue #577


cnk:
I think this automated page comparison should be visible to everyone. https://percy.io/cd0fd004/Hack-for-LA-Website/builds/6781468

PR #722, opened 2020-09-07
Hack Nights Page-issue-403

leonelRos:
Hey @ExperimentsInHonesty , here is the PR to the page please feel free to check it out and find the errors with the vectors. I will still try to figure out my error. @ye-susan  and I worked on it last Thursday but we could not figure it out. This is the error I am getting:

1. The page looks fully complete.image

2. After going to the mobile version and clicking the arrows for the drop-down and go back to the full desktop page we see the arrows in those two bubble cards.
image

To access the page add "/hack-nights-page" in the URL browser
Cheers


ye-susan:
Thanks Leo, I'll take a look at this tomorrow!

ye-susan:
Please see my initial review below: at a glance, most of the sections look really great, but here are just some notes about some items that we might want to fix before merging?


image

The header shows up like this for me, maybe we could ask to change the design to allow both paragraphs to be above the video on the mobile view, so that the paragraphs are in one column, video in another -- I'm hoping this may cause less complication.
A soft suggestion for the header container background color: reduce margin to none for header container and 'getting-started-page' class, this will allow the entire background to show as white, and the Events title can be adjusted individually

image

Some of the project mini-cards are being cropped off, when the browser window is not at max. The New Schools Today title is also overflowing outside the card 
image

leonelRos:
@ye-susan thanks for the feedback. I fixed the letters overlapping the box in the locations part. With the header, I was able to fix it without the margin but the title "Events" goes below the main navbar and when I tried to add some space between it the pink color comes back again. I am trying to figure out who added the color pink around the header. I checked my scss file and there is no pink color added in my file, so there must be other class components with the same name. Anyways, I will keep trying to figure this one out.

Screen Shot 2020-09-08 at 10 22 30 PM

Issue #703, opened 2020-08-19
Create project card for BallotNav


ExperimentsInHonesty:
### Overview
Provide collateral for the HackforLA website

### Action Items
- [x]   Gather items
  - [x]  600 x 400 image
     - [x]  Alt image text
  - [x] 1500 x 700 hero image (please do not put project title on hero image)
     - [x]  Alt image text
  - [x]  Name of project
  - [x]  A blurb about your project
  - [x]  Links (github, slack channel url, Testing Site, Live Site, etc.)
  - [x]  Any resources for a Getting Started link (a link to a wiki, readme(s) or both)
  - [x]  What you are looking for skills wise
  - [x]  Partner(s)
  - [x]  Location
  - [x]  Status
- [ ]   Add project Card to website

### Resources/Instructions
See project cards on the hackforla.org website for examples

Arjayellis:
600 x 400 image
 Alt image text
 1500 x 700 hero image (please do not put project title on hero image)
 Alt image text
 Name of project: BallotNav
 A blurb about your project: 
 Links (github, slack channel url, Testing Site, Live Site, etc.) 
- GitHub: https://github.com/hackforla/ballotnav
- Slack: https://hackforla.slack.com/archives/C0191KE1XB7 
 Any resources for a Getting Started link (a link to a wiki, readme(s) or both)
- Readme: https://github.com/hackforla/ballotnav/blob/master/README.md 
 What you are looking for skills wise
- Business Analyst (Category: BA)
- UI Designer (Category: UI/UX) 
- UX Researcher (Category: UI/UX) 
 Partner(s): Code for America, Other Brigades 
 Location: Remote
 Status: Active

Arjayellis:
@myastark I wrote a draft for the blurb on the project card. Would love your expertise before submitting. 

BallotNav aims to support absentee voters by addressing limitations caused by conflicting information from the media, hard to navigate government websites, and statewide regulations, with an efficient way to get accurate information about where, when and how to deliver your ballot in person.

This will take dedicated partners across the country to keep this resource up to date by daily monitoring of Secretary of State and county registrar websites across all 50 states and 3,007 counties through November 3rd (X number of days left). The end result will be a clear path to action for those who need it most.

This task is nearly impossible, but no one is coming. It’s up to us to save our democracy, and we can do it.

Arjayellis:
New draft based on @myastark suggestion. 

[...]

PR #724, opened 2020-09-09
Ballot nav card

Excursionpreneur:
Created a project card for BallotNav


KianBadie:
I see that the id for this project card is "11635254". However, the GitHub data in the [github-data.json](https://github.com/hackforla/website/blob/2785705bc5a9b289fd1d872fbbb6afc518605462/_data/github-data.json#L11959) file has the project id as "288594239". And I think this is the repo that @ExperimentsInHonesty just tagged with "hack-for-la" yesterday 9/14. If we wanted to show the contirbutors from the data I linked, we would have to change the project id to 288594239. I can do that myself too and merge (if there is already a visual confirmation form stakeholders). 

@ExperimentsInHonesty @Excursionpreneur. Was the given repo id from the stakeholder? Is there a specific repo they want us to pull data from besides the one that is linked by their project?

.gitignore
```diff
@@ -13,6 +13,7 @@ dist
 .jekyll-metadata
 .DS_store
 .vscode
+.idea
```
drubgrubby: Can you explain this change to .gitignore? I'm not sure I understand how it relates to the card.

_projects/ballot-nav.md
```diff
@@ -0,0 +1,35 @@
+---
+identification: '11635254'
+title: BallotNav
+description: BallotNav is a project of Hack for LA. Hack for LA is a brigade of a Code for America that exists to connect citizens with civic rights and responsibilities. BallotNav helps those who are concerned about using USPS for their mail-in ballots find their local drop-off points and their available days/times. The app's main features include locating the drop-off points closest to the user, and displaying what dates and times they are available; pulling in this information as it becomes available from county election supervisors; and connecting with users' social accounts to encourage dissemination of the tool and information.
```
drubgrubby: I see a version of a description/blurb in the issue, but it's different from what's written here.  This text clearly came from somewhere, but since I don't see it in the issue, I'm not sure how to review if it is correct.  What's the source of the description text? 
Here's what's in the issue: 
https://github.com/hackforla/website/issues/703#issuecomment-678710811

_projects/ballot-nav.md
```diff
@@ -0,0 +1,35 @@
...
+image: /assets/images/projects/ballot-nav.png
+alt: 'Map of the United States for Ballot Nav'
+image-hero: /assets/images/projects/ballot-nav.png
```
drubgrubby, line 7: This is not completely clear in the issue, but it seems like the last comment is that the picture of the boy dropping the ballot into the box is going to be on the card, and the map on the hero: https://github.com/hackforla/website/issues/703#issuecomment-679454433

drubgrubby (changes requested):

drubgrubby:
Overall this card looks great.  I have a couple of questions in my review above that may just relate to my not understanding something (which is not uncommon). 

One other note: if you start your PR with "Fixes #703" then it will close the issue when the PR is merged.

drubgrubby:
One other thing, which doesn't relate to the card as the meeting times aren't on the card, but may relate to the project id issue that Kian mentioned above, under the meetings times it says "undefined", at least when I look at it locally on my machine.  I see it on your screen shot, so it's probably not just my local machine.  Just mentioning it here so we can figure it out before merging.
Screen Shot 2020-09-14 at 5 33 12 PM

.gitignore
ExperimentsInHonesty: Please remove this file

_projects/ballot-nav.md
```diff
@@ -0,0 +1,28 @@
+---
+identification: '11635254'
```
ExperimentsInHonesty: This should be 288594239

.gitignore
KianBadie: @drubgrubby @ExperimentsInHonesty Ah, I should have documented this somewhere, my apologies. I was also curious about this file. I asked @cnk and she said it was because @Excursionpreneur was using the IDE [IntelliJ](https://www.jetbrains.com/idea/). We already had `.vscode` in the gitignore for the VSCode IDE. `.idea` is the equivalent for IntelliJ. Should we put that in a different pull request? I think they added it because some unnecessary files were initially being added tho this commit.

KianBadie:
@drubgrubby Good catch with the meeting times. @ruben1s and @cnk have done some work with the VRMS api, which I think is what the meeting time data depends on. Maybe we can talk about it more at tonight's meeting.

.gitignore
Excursionpreneur: I added .idea to the .gitignore file because they were being added when I was trying to push just the ballot-nav.md. And yes, I am using IntelliJ (PHPStorm).

PR #732, opened 2020-09-13
Reformatted current-projects template

cnk:
Someone's IDE had reformatted the IRB tags in this file. I am actually surprised this continued to work.


KianBadie (approved):
This looks good to me. Some of these I would say were objectively needed for readability, and others I would say are stylistic preferences. Either way, both are good with me, and the page seems to be functioning as usual. Overall, this is a plus. Will merge

PR #758, opened 2020-09-30
Renamed hack-nights to events and (mostly) related cleanup

### LINKED ISSUE #748: hack-nights page name-change related cleanup
### Overview
We have changed ../website/hack-nights-page.html to ../website/hack-nights.html.  

We now have both: 
../_sass/components/hack-nights-page.scss 
and
../_sass/components/_hack-nights.scss

It appears that some of the tags from hack-nights.scss are used on pages other than hack-nights pages. 

We also have both:
../website/hack-nights.html
and
../website/_includes/hack-nights.html

### Action Items
These files need the Code Zamboni(tm) - someone to go over these pages and clean them up.  We should get rid of resources that are no longer in use and consolidate the remaining resources. 

My suggestion:
1) See what styles on _hack-nights.scss are still in use and move them to the appropriate .scss page (most likely hack-nights-page.scss
2) Once all of the styles on _hack-nights.scss are accounted for, delete that page.
3) Rename hack-nights-page.scss to _hack-nights.scss

As far as the duplicate, hack-nights.html pages... I'm not sure that ../website/_includes/hack-nights.html is still in use.  If not we can delete this page.  If it is, determine if there is a problem with two pages having the same name (it seems like a bad idea to me), and, if yes, take appropriate action.

drubgrubby:
Fixes #748 

This main part of this was renaming hack-nights.html to events.html.

In addition, there was some cleanup to related files:

Add underscores to .scss files for consistancy
Remove undescore from _main.scss 
(cnk suggested the above two changes to keep naming convention consistent)
Change hack-nights-page.scss to _events.scss
Delete ../website/_includes/hack-nights.html
Moved a couple of styles from hack-nights.scss to events.scss
Added events.scss to main.scss


KianBadie:
@cnk asked me to do a review visually on this. I did a pretty shallow visual review because I did a pretty thorough-ish review of the css classes and came to the conclusion that much shouldn't change visually because most of the classes were unique to the page on not used anywhere else. Everything looks good to me :+1: 

However (not because of your work @drubgrubby but I think these notes stem from things that were already existing on files related to the page):
- `.meetings-message` was used/defined in `project-meetings/html.scss`, however they both use the same exact rules so no problem there
- `active::after` is used in 2FA guide page and `getting-started-page.html/scss`, however the rules on the `events` page just reapply the same exact rules to mobile, so no problem there either
- I know `.section-hack-nights` and `.btn` got moved to `events.scss`, however it is effectively imported in the same order, as `events.scss` is imported where these classes used to be imported. So no problem there as well as far as I can tell. 


@cnk @DanielekPark I did notice a bug (that I'm pretty sure was inherited and not from your work @drubgrubby), but because `active::after` is defined in `getting-started-page.scss` for desktop views, it applies to this page as well. Which means, if you click the "Meeting Events" header under "Our Events", it will add a drop down arrow to the left of it. Click it again and it makes the content disappear. This is already on the main site, so again, it's not from this work.

Good work @drubgrubby

PR #764, opened 2020-10-02
MVP events mobile page

### LINKED ISSUE #403: Create a Hack Nights Page
### Overview

We are going to move to a model where the hack nights have their own pages when we go back to meeting in person. In the meantime, we need a hack nights page, that says what the status of the hack night is, while maintaining our promotion of the amazing had nights we had.

Details of what the page should contain:
- The images that used to exist on the homepage
- list of projects and icons of those projects by location
- Maybe - links to our last in person meetings via meetup - maybe
- Encouraging people to signup for meetup

### Action Items
- [x] Review the other prototypes
- [x] Design low-fidelity wireframe that mirrors that design but is specific to hack night information
- [x] Draft copy for Hack Nights page
- [x] Revise and proof copy for page
- [x] Design high-fidelity mockup
- [x] Insert copy into finalized mockup
- [ ] Implement new Hack Nights page

### Resources/Instructions
Copy: https://docs.google.com/document/d/1wVbTCuALfUs02qVahkwfAoZ0BSQyYcp_TCbvja0hSrY/edit

[Lucid Chart](https://www.lucidchart.com/pages/)
Prototype the Projects page #156
Prototype the Project-specific page #157

drubgrubby:
Fixes #403 

This makes the events mobile page look much better than it does currently.  I listed it as an MVP because it is not very refined, and I haven't done a lot of cleanup on the css classes, and other things that I would normally do.  I just got it to where it looks basically like the figma. 

Given the relentless nature of the Jewish holidays at this time of year I don't have time to do all of the refinement and cleanup I would normally do. Given that this page is live and the current mobile display is...not very good, I'm proposing that we merge this so it is at least acceptable, and then - if it is deemed necessary - we open a new issue called something like "clean up events page mobile" and I will get it cleaned up. 

If you think it's okay the way it is, then we can just get on with our lives and on to the next project. 

If you think we should wait to merge it until it's perfect(ish) then I'll get to it next week. 

Current events page mobile: 

Screen Shot 2020-10-02 at 9 56 41 AM

How it would look after this merge: 

Screen Shot 2020-10-02 at 9 45 07 AM

Note: I think this design is problematic as soon as you start to make the page size smaller than full screen and the text on the right starts to get narrow and push way down below the video display - this is the display that you see on a tablet.  Design things are not my decision to make, but I might suggest that we make the way the mobile looks in this merge start at tablet size, so the video is in the middle of the two text blocks rather than to the left of them.  Just my two cents.


ExperimentsInHonesty (commented):
It looks good visually to me and I agree with incremental fixes to what is there now.  If Kian says code quality looks good, then it can be merged.

KianBadie:
@ExperimentsInHonesty @drubgrubby sorry for the delays, but I will be able to give this pull request a review tomorrow. I was considering giving this a quick review before the meeting (because I can not do a review after the meeting today), but I figured doing it tomorrow and looking closer at what the code is doing would provide a better review.

events.html
```diff
@@ -15,21 +15,34 @@ <h2 class="header-desc">
           toward using the web as a platform for local government and community
           service.
         </h2>
+        <!--This is hidden when full screen and shown when mobile-->
+        <div class="fullscreen--hide-video" style="width: 100%">
+          <div style="position: relative; height: 0; padding-bottom: 56.25%">
+            <iframe
+              class="video-size"
+              src="https://embed.ted.com/talks/jennifer_pahlka_coding_a_better_government"
+              width="854":w3
```
KianBadie: What does the `:w3` statement do? I tried doing a quick search online but could not find information on it. It obviously works and isn't breaking anything because it passed the visual. But I am wondering what the affect of it is?

events.html
```diff
@@ -15,21 +15,34 @@ <h2 class="header-desc">
           toward using the web as a platform for local government and community
           service.
         </h2>
+        <!--This is hidden when full screen and shown when mobile-->
+        <div class="fullscreen--hide-video" style="width: 100%">
```
KianBadie: I noticed that the component that shows the video seems identical to the component that shows on fullscreen. Is what I observed correct, or are there differences I did not catch? Because if they are identical in styling and structure, then I think we could just have one component for both fullscreen and mobile.

events.html
```diff
@@ -15,21 +15,34 @@ <h2 class="header-desc">
...
+              height="480"
+              frameborder="0"
+              scrolling="no"
+              allowfullscreen
+            ></iframe>
+          </div>
+        </div>
 
-        <p id="head-content" class="hacknights--head-content">
+        <div id="head-content" class="hacknights--head-content">
```
KianBadie: This one is just more so a curiosity question, did making this a div tag just make it easier to work with? Because if it does, that is something I want to keep in mind.

_sass/components/_events.scss
```diff
@@ -458,6 +459,30 @@
...
+  }
+
+  //This hides the fullscreen positioned video for mobile
+  .mobile--hide-video {
+    display:none;
+  }
+
+  .hacknights--head-content {
+    margin: 60px 0 0 0;
+    text-align: initial;
```
KianBadie: Small nitpick, but I do not think the additional `text-align: initial` is needed since that rule is defined in the non-mobile class definition for `hacknights--head-content`. This line doesn't do any harm functionally but I think it would be good to keep the amount of styling rule declarations at a minimum so code can be read conveniently. Thoughts?

KianBadie (changes requested):
@drubgrubby Good job on the changes, it looks a lot better on mobile :+1:.  I left some review comments on the code structure wise. I also know this got the visual confirmation, so I didn't want to make these points as request changes, but wanted to see if they were significant enough that @ExperimentsInHonesty  wanted to integrate a change for them. But for me, **since I was just doing the code review, responses to the review comments is all I need to give this a pass (with a response as well to the inline styling rule note I put below, since it is code related)**:

- On landscape for phones, the bottom head content has a sizable space between the video (screenshot below). Would we want to implement a styling rule that lowers those margins on mobile?
- On large phone on landscape, the text get's very tall and thin like you mentioned with the tablet views @drubgrubby. Would it be easy/not too much to add to extend the styling rules that control that text on mobile to fit the width of large phones on landscape? I think the [toolkit styling](https://github.com/hackforla/website/blob/142732277d65420913dd2ccb0c37e4f53a9bd145/_sass/components/_toolkit.scss#L140) has some good examples of custom queries for screen width.
- There are some inline styling rules in the components (that look like they were inherited rather than created by you) related to this task, but I would opt to just have that in a "clean up events page" issue so @drubgrubby can enjoy the holidays. I don't want to pile on additional things. Unless you think it is important @ExperimentsInHonesty to be done with this issue and if you don't mind @drubgrubby.

Screenshot from 2020-10-05 17-24-08
pixel2xl_landscape

drubgrubby:
I addressed the first three comments as follows: 
1) I removed the inline styles and changed the margin and it seems to have solved the extra space problem.
2) I changed the size at which the video location changes relative to the text to get rid of the long, thin line of text on the right.  Now the layout changes at 960px width. 

Just before the change (as you're getting smaller): 
Screen Shot 2020-10-06 at 4 41 21 PM

Just after the change: 
Screen Shot 2020-10-06 at 4 39 58 PM


3) I removed those inline styles.  As far as I can tell they didn't do anything, so no one should miss them.

events.html
drubgrubby: I have absolutely no idea.  I tried to do a quick search as well, but there's so much "w3" stuff around that I couldn't find anything either.  It was already there; it was not something I added.

events.html
drubgrubby: The divs that contain the video are identical.   The reason I have it there twice is that it was the only way I could figure out to have it be two different places relative to the text on the page - one before the text and one in the middle of the text.  If I were writing the page from scratch I might have done it differently (though I can't think how at the moment, given the design), but as the full size page was already working and I was just trying to get the mobile to look better, I did it this way.  I'd be delighted to have the component there only once if you have a good way to do that.

events.html
drubgrubby: As I recall, when I inherited the page it was a `<p>` tag.  As I was moving things around it annoyed me because there didn't seem any reason for it to be a `<p>` rather than a `<div>` like all of the others, so I just changed it to make it the same.  It was neither easier nor more difficult to work with, it was just visually pleasing to me.

_sass/components/_events.scss
drubgrubby: Agreed.  This was one of those "I'm in a hurry. It works. I'll just leave it." things.  I have removed that tag.

events.html
drubgrubby: Okay, that's weird.  I just looked at the changes and it does look like I did that, but I don't think I did, unless I was sleep coding.

KianBadie: Resolving conversation as  `w3` has been removed. Thank you!

events.html
KianBadie: Ah I see, I was thinking in an isolated way were I was only think about the component itself. However, I did not even realize it was in a totally new place in the DOM. Nice solution, I'll resolve this conversation :+1:

_sass/components/_events.scss
KianBadie: Thank you :+1: I will resolve this conversation.

KianBadie:
Thank you for the updates @drubgrubby :+1: unfortunately, I will not be able to get to re-reviewing today as I thought. However, tomorrow I will make it a priority

events.html
KianBadie: For sure, I think I am actually going to make a quick edit to revert this back to a `<p>` tag as it provides [semantic](https://www.quora.com/What-is-the-difference-between-a-p-tag-and-a-div-tag) meaning.

KianBadie:
@drubgrubby I added some commits to add back in those inline sty lings in the form of a class. It looks like the styling were adding some minor changes when in full-screen. I also changed that `<div>` tag to a `<p>` tag because it provides semantic meaning, and I think it is better practice to use elements with semantic meaning. I committed these myself because it was small, and I wanted to save you the trouble.

Other than that, I noticed on the smallest screen of iPhone 5, the display was not working (screenshot below). This is currently existing on the main site as well, so I do not think it is something from the code @drubgrubby added. The code that improves the mobile view the other screen sizes currently makes sense and look good. So I will leave it up to @ExperimentsInHonesty if she wants to merge with the iPhone 5 issue.

Screenshot from 2020-10-08 15-35-23

Issue #162, opened 2019-10-23
Develop Wins page


alecg123:
### Overview
Wins Page will document our project successes as well as individual team members 'growth+development+and achievements'; including new hires and skills.
 
One of our HackForLA Website Update Objectives is to “Present a clear and compelling value proposition for participants in projects.” Essentially what can a volunteer get out of volunteering for Hack for LA? 

### Action Items
- [x] Product
   - [x] Brainstorming what this Wins Page might look like.
- [x] UI
   - [x] create a page layout that has the following
      - [x] link to "Share your wins" form
      - [x] testimonial layout
         - [x] name
         - [x] linkedin image and link
         - [x] checkboxes 
         - [x] quote
- [x] Product
   - [x] Create google form
____
- [ ] Redesign the cards to include badges on the bottom. And some way of searching through the cards
- [ ] Create a slack bot that informs the group when there is a new win.
- [ ] Change the order of the info on the card to have the one they type out first.
- [ ] Add more lines of text on each card.

### Resources/Instructions
see this comment for details about the type of checkboxes https://github.com/hackforla/website/issues/162#issuecomment-636576691

Secrets issue https://github.com/hackforla/website/issues/98

### Important links
- [description of slack bot](https://github.com/hackforla/website/issues/162#issuecomment-739546005)
- [google sheet with all the wins](https://docs.google.com/spreadsheets/d/1tj6eQlVLFgmskoXouVt87POxYVwR1yWT2vOSQEPyDtg/edit#gid=1106372497)
- [figma](https://www.figma.com/file/0RRPy1Ph7HafI3qOITg0Mr/Hack-for-LA-Website?node-id=1284%3A102)

Mr/Hack-for-LA-Website?node-id=1284%3A102)

ExperimentsInHonesty:
@alecg123 said: We might advertise on this page: 

- volunteers finding new jobs
- the number of people affected by a project win
- new commits added to team members’ GitHub account
- experience in team settings
- project management opportunities, etc.

ExperimentsInHonesty:
I got a new job
I produced something for my portfolio
I improved my linkedin
I learned how to work better on a team
I increase the number of commits on my github profile
I learned a new language
I became proficient at github
I setup 2FA
I worked on an enterprise project
I worked on a project that will help the people of Los Angeles
I worked on two projects
I taught or mentored another person on my team
I taught or mentored a person on another team
My team delivered software to a stakeholder
My team launched our MVP
Other _______________

Tell us about it:

[...]

PR #775, opened 2020-10-11
Wins refactor - adds display approval process

jbubar:
#162 
Added column in google sheets for displaying approval
Screen Shot 2020-10-11 at 10 38 26 AM

Added logic to exclude win inputs that are not approved to be displayed
screencapture-0-0-0-0-4000-wins-html-2020-10-11-10_34_25


rblaz001:
Looks good, verified that it is working as intended.

PR #779, opened 2020-10-14
AdBlocker is blocking navigation and donate button

### LINKED ISSUE #616: Test Adblocker on HfLA Site and Footer
### Overview
We need to ensure that adblockers are not blocking our code from showing up to ensure that visitors to the site are able to see content on the footer. 

### Action Items
Developer
- [x] Install Adblocker (see link below)
   - [x] Try to replicate issue by turning blocker on (no special settings)
- [x] If issue is confirmed, check to see if we can update code to avoid this happening again 
- [x] If code change is not possible or does not resolve, add the social information back to the header/navigation of the site

### Resources
[Adblocker Google Chrome extension](https://chrome.google.com/webstore/detail/ad-blocker/kacljcbejojnapnmiifgckbafkojcncf). 

### Visuals
Screenshot of what it looks like here: 
Screen Shot 2020-07-19 at 3 43 46 PM

drubgrubby:
Fixes #616 (mostly)

**tl;dr:**
- By removing the word "social" from css tags, the navigation and donate buttons on the header and footer are visible. 
- Ad blockers are all different.  Some of them still block some of the social media links (twitter, facebook, etc) in the footer. 

**Full explanation:**
All ad blockers seem to be a little bit different.  The main one that I had problems with was 1blocker, which (as I understand it) just runs on Safari.  I also had some problems with AdBlock on Chrome.  

It seems that the issue falls under the heading of "Annoyances".  I guess these are things that they think will annoy people, which I find kind of annoying. 

In the 1Blocker docs: 
"Block Annoyances (39.7K rules)
Block social media widgets, cookie notices, share bars, crypto-miners, etc."

One part of their blocking "social media widgets" is that it seems to block anything that has a css tag that has "social" in the title.  By renaming the css classes with the word "social" in it I was able to avoid blocking of the header navigation and the donate button.  

I changed the following tags: 
social-links tag to header-links
social-list to header-list
social-header to footer-header
js-social-link to js-inline-link
js-social-link-meetup to js-inline-link-meetup

It seems that they also have a rule that specifically blocks by name "twitter", "slack" and other social media names. Because of this most of the social media links in the footer were blocked.  (Not, it seems, meetup).  I'm not sure if there's a good way to link to twitter without having "twitter" in the url.  (As I'm writing this I suppose you might get a tiny-url and rename the image something that doesn't have "twitter" in it.)  

**Since this is not all ad blockers, and the current PR makes the navigation and donate buttons visible, I suggest we push this now and then decide if dealing with the footer links is worth the trouble.**  

Most of the comments about trying to get around this said some version of "if this is what the user wants, we should respect that".  I basically agree with that. 

**Screenshots**
This is a desktop browser, in chrome, with AdBlock after changes
Screen Shot 2020-10-13 at 4 40 07 PM

Header on iPhone before changes with 1Blocker on
UNADJUSTEDNONRAW_thumb_259e

Footer on iPhone before changes with 1Blocker on
OWyO32rUR7u1TEX+nQka%w_thumb_259d

Header and Footer on iPhone after changes with 1Blocker on
UNADJUSTEDNONRAW_thumb_259f

Header and Footer on iPhone after changes with 1Blocker off
UNADJUSTEDNONRAW_thumb_25a0


drubgrubby:
The footer on vera.org/events is being blocked in the same way as hfla.

https://github.com/hackforla/website/issues/616#issuecomment-696294935

Blocked
Screen Shot 2020-10-13 at 6 09 39 PM

Unblocked
Screen Shot 2020-10-13 at 6 12 01 PM

PR #783, opened 2020-10-16
Make internal links relative

cnk:
It is really hard to do a good job reviewing changes if you suddenly end up on the live site instead of staying on the local copy

### DIFF
--- a/_guide-pages/2FA.html
+++ b/_guide-pages/2FA.html
@@ -13,7 +13,7 @@ display: true
 category: development
 svg: svg/2FA.svg
 alt-text: 2fa-image
-provider-link: 'https://www.hackforla.org/guide-pages/2FA'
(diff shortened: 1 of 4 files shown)


KianBadie:
Haha, thank you @cnk. Can you please explain to me why relative links are preferred to absolute links? Is it because absolute links can change (ex: `hfla,org` -> `hackfla.org`), so a relative link will work around this?

cnk:
When in production, it isn't a big problem to have the full urls. And if we change the public url, we can just run a search and replace and then publish.

The reason I don't want http://hackforla.org in the urls is because when I click around to check things locally, I suddenly end up on the live site and can't properly check the changes.

KianBadie:
@cnk Ah I see, well that seems like a compelling enough reason in itself. Thank you for the explanation :+1:

PR #784, opened 2020-10-16
Fix toolkit link in main navigation

cnk:
This needs to be /toolkit so we can get back from individual guide pages


KianBadie:
Thank you for the catch @cnk

PR #788, opened 2020-10-17
Update styling for toolkit header.

### LINKED ISSUE #780: Standardize Toolkit Page Header CSS
### Overview 
The Toolkit page's header is currently inconsistent with other similar page headers. Adjust CSS (padding and text) to conform with rest of HfLA site.

### Action Items 
- [ ] Follow instructions outlined in [this Figma file](https://www.figma.com/file/0RRPy1Ph7HafI3qOITg0Mr/Hack-for-LA-Website?node-id=3214%3A34) to standardize Toolkit page header
     - [ ] Adjust text size, font weight, line height as noted
     - [ ] Adjust padding as noted

### Resources/Instructions  

[Toolkit Page Edits Instructions on Figma
](https://www.figma.com/file/0RRPy1Ph7HafI3qOITg0Mr/Hack-for-LA-Website?node-id=3214%3A34)[Current Toolkit Page
](https://www.hackforla.org/toolkit)[Current Getting Started Page (for Comparison)](https://www.hackforla.org/getting-started)

KianBadie:
Fixes #780 

I'll be specific in what I did to match/respond to the specificity of the directions.
- Made title of page exact same as getting started by removing toolkit page specific class (now it just takes styling rules from h1 tag just like getting started page).
- Made font size, max width, padding, and margin of header paragraph the same exact values as the getting started header paragraph. Only tangible difference between the two pages for this part is the text align (I kept it left for the toolkit as that was the original request). I didn't make a generic class for partly that reason and partly that I did not have the time for it today. And I also didn't know if I should start that process.
- Reduced the space on the top and bottom of the header. Not the same values as recommended in the figma, but it is the same as the getting started page (which I felt like was the goal of the issue). They are both computed to 40px away from the nav bar and the following section (content for the getting started page, category banner for toolkit page).
- "Guides We Made These!" and "External Resources" font size changed to 1.75rem (which equates to the request 28px). I kept this in rem because the toolkit used much more `rem` units than `px` units, and I wanted to keep that theme as much as possible. I kept the `px` units in the header paragraph so it can match the getting started class definition more closely, and the componentization/genericizing of the 2 classes more apparent.
- Made Category banner 1.5rem as requested (which equates to 24px)

On iPhone 6/7/8 landscape (and other width lengths around that), it looks like the title of the page is the same font as the paragraph. However, it is the same behavior on the getting started. I posted a screenshot in it just in case it was considered an important thing to consider.

I did not post many screenshots and finished this issue in a bit of a hurry. So if you wanted to have an in depth look please clone the branch to local and self test. If more screenshots or changes are desired, I can add those tomorrow. I opened this pr just in case this issue was urgent and needed at least something.

**Edit**: I also realized that I didn't make the computed px distance between the header paragraph and the page title/page image 70px as requested. However, again, it is the same as the getting started page. Also, a difference between the getting started page is that on mobile the font of the paragraph does get smaller. So there is at least one difference in behavior.

# Screenshots
## Desktop
desktop
## iPhone 6/7/8 Vertical
iphone678top
iphone678bottom
## iPhone 6/7/8 Landscape
iphone678landscapetop
iphone678landscapemiddle
iphone678landscapebottom

### DIFF
--- a/_includes/head.html
+++ b/_includes/head.html
@@ -12,7 +12,7 @@
   <title>{% if page.title %}{{ page.title | escape }} - {% endif %}{{ site.title | escape}}</title>
   <meta name="description"
     content="{% if page.description %}{{ page.description | strip_html | strip_newlines | truncate: 400 }}{% else %}{{ site.description }}{% endif %}" />
-  <link rel="stylesheet" href="{{ '/assets/css/main.css' | absolute_url }}" />
+  <link rel="stylesheet" href="{{ '/assets/css/main.css' | relative_url }}" />
   <link rel="canonical" href="{{ page.url | absolute_url }}" />
   <link rel="shortcut icon" type="image/x-icon" href="{{ 'favicon.ico' | absolute_url }}" />
 
--- a/_includes/head.html
+++ b/_includes/head.html
@@ -12,7 +12,7 @@
   <title>{% if page.title %}{{ page.title | escape }} - {% endif %}{{ site.title | escape}}</title>
   <meta name="description"
     content="{% if page.description %}{{ page.description | strip_html | strip_newlines | truncate: 400 }}{% else %}{{ site.description }}{% endif %}" />
-  <link rel="stylesheet" href="{{ '/assets/css/main.css' | relative_url }}" />
+  <link rel="stylesheet" href="{{ '/assets/css/main.css' | absolute_url }}" />
   <link rel="canonical" href="{{ page.url | absolute_url }}" />
   <link rel="shortcut icon" type="image/x-icon" href="{{ 'favicon.ico' | absolute_url }}" />
 
--- a/_sass/components/_toolkit.scss
+++ b/_sass/components/_toolkit.scss
@@ -44,6 +44,7 @@
 .suggest-guide-group {
     display: flex;
     justify-content: space-between;
+    align-items: center;
     padding-left: 3rem;
     padding-right: 3rem;
     text-align: center;
@@ -227,7 +228,6 @@
     }
 
     .suggest-guide-group {
-        align-items: center;
         flex-direction: column;
 
         h2 {
--- a/_sass/components/_toolkit.scss
+++ b/_sass/components/_toolkit.scss
@@ -26,6 +26,7 @@
     }
 
     &__hero-image {
+        margin: 33px 0px;
         max-width: 100%;
     }
 }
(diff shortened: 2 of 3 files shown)


KianBadie:
I added some more small updates to try to make this more identical to the getting started page. I have not done a thorough comparison on mobile, but I did add a gif below trying to show how the page looks at different widths (I did not show the difference in the "Guides (We Made These!)/External Resources" headers, but it does stay at the requested font size on all widths).

desktop
widthdemo

drubgrubby (approved):
@KianBadie - These changes look great and this page seems to match the Getting Started page very closely. 

- Nice that you stripped back to the H1 heading.  This sort of thing will help make this a much easier site to code from a CSS perspective.  

- I agree with you leaving the left align on the header text.  Not only does it match the figma and look better, but it's the same on the AboutUs page that I'm working on. 

- The main difference I notice between this and the Getting Started page is that the text size changes on the toolkit page when you get below mobile  ($bp-below-mobile), and it doesn't change on the Getting Started page.  I think it looks much better on the toolkit page and recommend, merging this and (at some point) making changes to the Getting Started page. 
Screen Shot 2020-10-19 at 9 55 46 AM
Screen Shot 2020-10-19 at 9 55 56 AM

- The only thing that my eye bumps up against on this page is that the bottom of "Guides (We Made These!)" is not aligned to the bottom of the "Suggest a Guide" button; they are aligned at the top.  Because the text and the button are not the same height, there's an uneven space on the right and left side below the text (on the left) and the button (on the right). It's super-minor and is also true on the figma (to a lesser extent).  I don't think it's worth holding this up for, but i'm just putting it out there in case it's a quick-fix, and fixing it doesn't put something else out of whack. 

Screen Shot 2020-10-19 at 9 59 50 AM
 
(See the bottom of the button still showing after the text is gone).

KianBadie:
@drubgrubby Thank you for the thorough review and responses to the details. Also, for you recommendation of centering the text/button in the "Suggest a Guide" heading: It was indeed a quick fix. Since it is a minor change, I imagine there would be little debate on including that change, so I went ahead and centered those 2 elements you mentioned. From this screenshot, it appears that the text on the left is lower (using the button on the right as a reference point) than it was before. That is because the elements are centered on the horizontal axis. (Screenshot below)

I will go ahead and merge this since @ExperimentsInHonesty request it, and I imagine this minor fix I added is non-controversial.
Screenshot from 2020-10-19 17-09-56

PR #792, opened 2020-10-21
Fixed misaligned chevron

### LINKED ISSUE #725: Chevron Misalignment - Getting Started Page
### Overview
One of the chevrons on the lower panel does not align correctly. The chevron positions itself below the line under certain mobile views.  

### Action Items
Find out what is causing the chevron to be misaligned under certain mobile views. 

### Image Example
Old

ruben1s:
Fixed misaligned chevron. addresses issue Fixes #725

### DIFF
--- a/_sass/components/_getting-started-page.scss
+++ b/_sass/components/_getting-started-page.scss
@@ -91,6 +91,10 @@
   // NEW MOBILE FRIENDLY RULES
   @media #{$bp-below-mobile} {
     padding: 30px 15px;
+    p.accordion {
+      display: inline-block;
+      width: 100%;
+    }
   }
 }
 
@@ -136,10 +140,6 @@
   }
 }
 
-p.accordion {
-  display: inline-block;
-  width: 100%;
-}
 
 .svg-accordion {
   float: right;
--- a/_sass/components/_getting-started-page.scss
+++ b/_sass/components/_getting-started-page.scss
@@ -136,6 +136,11 @@
   }
 }
 
+p.accordion {
+  display: inline-block;
+  width: 100%;
+}
+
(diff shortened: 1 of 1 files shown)


drubgrubby (approved):
I approve this for merging. This fix does what it's supposed to and fixes the problem without creating another problem somewhere else.  I wish that I could wax rhapsodic about this, that, and the other thing, but this seems to be a perfect simple fix that fixes the problem, so there's nothing much to say.

KianBadie:
@drubgrubby thank you for the reivew. Nice fix and work @ruben1s. Please post screenshots of fix in issue and change the wording in the initial comment of this pr to "Fixes #725" so that the issue can be automatically closed. 

At first I was questioning why the definition was in the `.getting-started-page-card`. But as @drubgrubby pointed out, I realize now the way you structured it makes it so that the this indeed only affects the paragraph text **with** accordion elements **only** in `getting-started-page-card` classes, so it really reduces the probability of it affecting other pages. Nice :+1:  

Since the card is bigger than it used to be (only on mobile), I think I would just like to double check and get visual confirmation from @alexandrastubbs or @ExperimentsInHonesty. I'm cool to merge if it gets a visual confirmation.

A side note outside the scope of this issue: the .accordion element is redefined for the 2FA guide page css, which I feel like is another reminder of the web of tangled definitions that exists on this site.  

Since I'm already here, I will attach an example screenshot myself for visual confirmation reference.
## Before
Screenshot from 2020-10-21 17-34-09
## After
Screenshot from 2020-10-21 17-33-59

alexandrastubbs:
Looks good @KianBadie ! Approved visually.

PR #820, opened 2020-11-30
Loading screen

### LINKED ISSUE #698: Dashboard Page Loading Time Message
### Overview
We need to add a message that appears when the dashboard page is loading so that users know the page is in fact working and it will just take a bit longer. 

### Action steps
- [ ] Development
   - [ ] ideate solutions for how we can add a message to the page while loading 
- [ ] Product 
    - [ ] review solution and source any needed visuals 
- [ ] Development
   - [ ] implement solution

### Resources/Instructions
[Dashboard page](hackforla.org/dashboard)

sena-ji:
image
This is the current loading screen that I have for the website. Please let me know if any changes to the design is needed.
Fixes #698


drubgrubby (commented):
This seems to work exactly as a loader should.  Nice work!

One question/thought: If I set it for a slightly slower connection while testing there's a quick flash of blue and then the page loads.  It's not clear that it's a loading page as it flashes by so fast.  I'm wondering if you can put a setTimeout or something on this so that if the page loads quickly it doesn't even run the pageloader script.

drubgrubby:
@sena-ji - I'm sorry I didn't notice this earlier.  This loader page only needs to go on the dashboard page, not the whole site.  https://www.hackforla.org/dashboard  The dashboard page loads a bunch of data in an iframe and it always loads slowly.  

If the method you used for testing when the page loads doesn't work, there's a link on the issue that might help: https://github.com/hackforla/website/issues/698#issuecomment-683454932

sena-ji:
Got some advice for this issue on Tuesday's developer meeting. For some reason, when the iFrame was "loaded" it would still be loading the documents contained in the page. Seemed like the loading page was ending early. So I added a timer to set a delay in hopes that the iFrame will fully loads it's documents before it ends. Also some of the styling for the loading page was conflicting with the rest of the page. I made some changes to fix the styling issue.

drubgrubby:
Nice work @sena-ji !   This took persistence to push through the various roadblocks that came up along the way. 

This seems to be one of those issues that doesn't have a perfect solution because we only have so much control over what's gong on in the iframe.  This feature, especially with the additional timer functionality, is so much better than it was before that I think it's good to merge.

PR #821, opened 2020-12-01
update wins data temporarily so that it will display

jbubar:
This pull request will be null after 4am tomorrow. 

It updates the wins-data.json file so that it displays all of the new wins! (for giving tuesday)(in case bonnie wants)

### DIFF
--- a/_data/wins-data.json
+++ b/_data/wins-data.json
@@ -132,7 +132,9 @@
       "",
       "6/1/2020",
       "I got a new job",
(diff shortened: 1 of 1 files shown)


jbubar:
Here is the screenshot of the site with the changes.. I also noticed that I had the cards in the wrong order. so I fixed that with another pull request #822 
screencapture-0-0-0-0-4000-wins-2020-12-01-11_44_49

PR #829, opened 2020-12-06
We need to filter out the 'hidden' projects before building our filters.

### LINKED ISSUE #657: Status dropdown - On Hold count/card-pull mismatch
### Overview
On the Status dropdown under "Current Projects" the number next to "On Hold" shows as "(7)", which I'm assuming is the count of how many cards == "On Hold".  However, when you select On Hold only three cards are displayed.  

### Screen Shots
e80L9XrGRc6wYlRHEf6mww_thumb_243e

cnk:
Because of the way Jekyll's where method works, that's easier if we match something rather than look for 'not' something.
So I replaced "hide: true" with "visible: false"  (and added visible: true to all the visible project cards)

Fixes #657

### DIFF
--- a/_includes/current-projects.html
+++ b/_includes/current-projects.html
@@ -16,7 +16,7 @@
         {% for status in status_list %}
           {% assign projects = site.projects | where: "status", status | sort: "title" %}
           {%- for item in projects -%}
-            {%- if item.hide != true -%}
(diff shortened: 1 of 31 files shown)


drubgrubby (approved):
Works like a charm.

PR #845, opened 2020-12-15
About-us page scss file cleanup

### LINKED ISSUE #842: About-us page scss cleanup
### Description
There are many classes in this file that would probably be better off being moved to different "master" .scss files so that they can be used more easily on different pages.  Additionally, the file could be somewhat better organized to so that future coders working on this page won't curse my name quite so much. 

### Resources
_sass/components/_about-us.scss

### ToDo

- [ ] Clean up scss file
- [ ] Check this off the master issue here -> https://github.com/hackforla/website/issues/835#issuecomment-740081272

drubgrubby:
Fixes #842 

This PR serves two functions: 
1) To clean up the scss file, clean up some comments, and in general make it more readable and useful. 
2) To begin to componentize the website by moving some of the generic classes to their own scss files. 

Also, while I was doing this I figured out a solution to a nagging problem I'd been having with the letter from the executive director heading, so I just fixed it before I forgot my brilliant solution. 

I apologize for how difficult the about-us.scss file is in the comparison here on the Pull Request.  Because I moved so many things around, and moved some classes off and then back, and realized that I'd accidentally deleted a necessary class so I had to copy it from a different branch...it looks kind of like my dog ate it.  I'd suggest just downloading it to your local machine and looking at it that way.  

But do what you will.


daniellex0:
@drubgrubby I downloaded this as suggested and everything looks good to me. It's consistent with what we discussed about componentizing the site and creating generic scss files for cards, containers, etc. 

I will likely make some changes down the line, but this is a great start and gets me in the right direction.

PR #870, opened 2020-12-27
Update 100-automations.md

mdothman:


Olivia-Chiong (commented):
Change roles for each member who was added.
Add UI/UX Designer to looking for

Issue #937, opened 2020-12-29
Update Project Profile: Lucky Parking


ahdithebomb:
### Overview
Please review the following information for your project and provide any updates 

### Project card
https://www.hackforla.org/projects/lucky-parking

### Action Items
Please check off everything that is ok the way that it is, and if not, please provide the updated information in the comments below:
- [x] title
- [x] description
- [x] image
- [ ] leadership
   - [x] Name 
   - [x] Role 
   - [x] Github Handle
   - [x] Slack Member ID
- [ ] links
  - [x] Github
  - [x] Slack
  - [x] Test Site
  - [ ] Demo Site
  - [ ] Site
- [x] looking
- [x] technologies
- [x] location
- [x] partner
- [x] tools
- [x] status

jbubar:
@gregpawin 
Any updates on the Demo site, and site links?
I am merging the other changes, When you send the site links, please @ me so that I can make a new issue

ExperimentsInHonesty:
This issue never got moved to the in progress, which means it skipped the weekly meeting update.
When the pull was done, the text ```Fixes #937``` was not added to the description and so it didn't close the issue
Then because the issue was not closed, evan unassigned himself from it and it remained open and in the prioritized backlog.

ExperimentsInHonesty:
@gregpawin I have vetoed the language change from 

> Visualization of parking data to assist in understanding of the effects of parking policies on a neighborhood by neighborhood basis in the City of Los Angeles.

to 

> Helping the city planners and the community make informed decisions about parking policies in the City of Los Angeles

Unclear why you want the language change it seems like a change of purpose for the team.  We need to talk directly about that if that is going to happen and it would need to be reflected in a new proposed version of the project one sheet, etc.

jbubar:
@ExperimentsInHonesty The issue is unfinished, because all the links were not provided. It was left open on purpose.  We were planning on waiting on a response from @gregpawin

gregpawin:
The demo site is http://www.luckyparking.info. But it’s currently under
development. Not sure if we necessarily want to publish it anywhere.

gregpawin:
We can keep the original language: Visualization of parking data to assist in understanding of the effects of parking policies on a neighborhood by neighborhood basis in the City of Los Angeles.

gregpawin:
Also, I'd like to update the Project leadership:
# Project Leadership:

Product Owner / Data Science Lead: Greg Pawin
[...]

PR #972, opened 2021-01-28
Lucky parking Issue 937

evancchen:
Updated Lucky Parking project
Issue #937


jbubar (approved):
This looks great!
I see that we are still waiting on some links from the issue, but this is ready to merge

PR #985, opened 2021-02-02
fix sticky nav stickyness on About Us page

### LINKED ISSUE #968: About us: Fix nav not sticking on resize
### Overview
When the page is loaded on mobile, in certain situations, and then expanded to desktop, the sticky nav doesn't stick where it's supposed to.   I believe this has to do with the calculation of the top of the page. 

Screen Shot 2021-01-24 at 8 03 00 PM

### Action Items
Fix it so it doesn't do that anymore. 

### Resources/Instructions
https://www.hackforla.org/about-us

drubgrubby:
Fixes #968 

The sticky nav was not sticking when the page was loaded in mobile view, and then expanded to desktop view.  This was because in mobile view, where the sticky nav was hidden, there was nothing to use to calculate the spot where it should stick.  

I tried several different ways of recalculating stickyNavTop, but in the end nothing seemed more reliable than just hard-coding stickyNavTop.  Not very graceful, but it works.

### DIFF
--- a/assets/js/about-us.js
+++ b/assets/js/about-us.js
@@ -3,8 +3,18 @@
 /**************************************/
 
 //Initialize and set defaults
-let stickyNav = document.getElementById("sticky-nav");
-let stickyNavTop = stickyNav.offsetTop - 72;
+
+/* Originally stickyNavTop was calculated with the following code: 
+
+stickyNavTop = stickyNav.offsetTop - 72;
+
+However, when the page was loaded in mobile view, where there was no stickyNav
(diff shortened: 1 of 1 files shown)


t-will-gillis:
You must be a member of the HFLA website team in order to create pull requests. Please see our page on how to join us as a member at HFLA: https://www.hackforla.org/getting-started.  If you have been though onboarding, and feel this message was sent in error, please message us in the #hfla-site team Slack channel with the link to this PR.

PR #1000, opened 2021-02-04
Resize get in touch

### LINKED ISSUE #946: Another css fix
### Overview
Another Css fix needed on the main page:
Screen Shot 2021-01-17 at 11 02 31 AM

### Action Items
- [x] check out the _includes/forms/contact-us.html file
- [x] check out the _sass/components/_contact-us.scss file
- [x] identify the class that you need to add some stylings to
- [x] add a some css to create space between the text and the buttons with screen width is below 580

### Resources/Instructions
REPLACE THIS TEXT -If there is a website which has documentation that helps with this issue provide the link(s) here.

- This is where we add any & all available resources. Do not shy away from adding things. If you think your audience already has it or might not need it, don't let that stop you. Add it anyway! It is very time effective to have all resources available in a single place.

evancchen:
Fixes #946 
image


evancchen:
Breaking point is 600px

drubgrubby (approved):
@evancchen - Nice work!  Your changes solved the problem and now the buttons scale all together as they should.

PR #1006, opened 2021-02-07
Wins badges

jbubar:
This pr reworks the wins page to make it look cleaner and also adds the badges on the overlay
#162


jbubar:
the number of columns change based on the width of screen. The css is not perfect, I probably need to consult our local css grid heavy weight @drubgrubby.
Also.. I made the modal scroll if it is too tall for the screen, it would be great to get feedback on that behavior. 

Screen Shot 2021-02-07 at 2 58 33 AM
Screen Shot 2021-02-07 at 2 58 57 AM

jbubar:
Spoke to bonnie, danielle and oskar, and they requested that I make some changes to the mobile view and change the styling for the overlay a bit

Issue #959, opened 2021-01-24
Make the Toolkit page's category-filtering banner functional


daniellex0:
### Overview
We need to make the banner for filtering categories on the Toolkit page functional so that the banner will have a purpose, and so that users can filter the guides by the category the are interested in (Development, Design, Project Management, Professional Development). Keep in mind there are a few guides right now, but many more to come.

### Action Items
Research
- [x] Determine how to categorize the guides
- [x] Research how to filter by categories (can use the homepage filter list for reference)

Front-End Development
- [x] Add 'All' category to the far left on banner, which will be the default on page
- [x] Make categories on banner clickable
- [x] Implement filtering function for Toolkit page guides 

### Resources/Instructions
[Toolkit page](https://www.hackforla.org/toolkit)
[Homepage project filters ](https://www.hackforla.org/)

Filter banner:

<img width="1400" alt="Screen Shot 2021-01-24 at 3 59 10 AM" src="https://user-images.githubusercontent.com/67400366/105630012-3efeea00-5dfb-11eb-84ab-3c478a0f7f09.png">

mcspach:
Update
 
Progress: 
-Researched embedded Ruby syntax.
-Looked over the homepage filter for an example.
-Added 'All' category tab
-Made the links clickable, (but they don't filter the results yet)

Blockers: None
(@daniellex0 Needs some css adjustments for med screen sizes after adding the 'All' tab. Am I okay to make minor css changes to banner and links?)

Availability: Later this week.
ETA: Sunday Feb 14

<img width="1251" alt="Screen Shot 2021-02-09 at 10 37 01" src="https://user-images.githubusercontent.com/38964454/107412381-384bc480-6ad5-11eb-8395-02ac6ec6f136.png">

daniellex0:
@mcspach Sure, feel free to make CSS changes! 

I'm also realizing the banner is pretty tiny in mobile size because apparently a mobile version of this page was never actually designed, and so the devs had to figure it out on their own. Do you want me to design a quick mobile version (perhaps two rows), or would you want to try to figure something out to fix how tiny it is in mobile?

mcspach:
@daniellex0 Okay, thanks.

I will take a look at mobile and see what I can work out.

mcspach:
Not available Saturday 2/13.

###PBNE
Still working out the exact code, but think I can solve it in one more sitting.
No blockers, just getting comfortable with liquid.
If not complete, I'll have a better update on Tuesday dev meeting.

[...]

PR #1034, opened 2021-02-13
Implement filter on wins page (#1030)

### LINKED ISSUE #1030: Filtering dropdowns on wins page
### Overview
create buttons to filter by Role and Team
Screen Shot 2021-02-09 at 7 04 16 PM

### Action Items
- [ ] 

### Resources/Instructions
[figma](https://www.figma.com/file/0RRPy1Ph7HafI3qOITg0Mr/Hack-for-LA-Website?node-id=1209%3A0)
[project page](https://www.hackforla.org/#projects)

akibrhast:
Fixes #1030 

This PR Implements the filtering system on the wins pages.
issue#1030


drubgrubby (changes requested):
@akibrhast - Really nice work!  I've just done a quick look, but the code looks really clean and the functionality looks like it works as it should.  

A couple of nips and tucks: 
1) At a certain point, on mobile, the "Team" dropdown floats off the screen

Screen Shot 2021-02-13 at 8 12 04 PM

I'm not sure what the solution to this is, probably that the size should just get smaller ('cause you can't really stack them (Can you?)).  This is the one that really needs to be fixed before we merge (in my opinion). 

2) The box on the dropdown is wider than the box before it drops down, and it pushes out over the bottom of the other drop-down: 

Screen Shot 2021-02-13 at 8 15 21 PM

It does that on the current frontpage, so maybe that's a feature and not a bug.  I find it sub-optimal, but...what do I know. 

3) On Firefox there's a weird thing that happens where, if I tell it to load an iPhone 6 (or anyother mobile size) screen, the drop-downs don't work.  If I drag the screen to desktop, and then drag it back to the exact same size, they work fine.  I'm not sure if that means anything, or if it's just a browser anomaly, so I'm just mentioning it. (There's a way to check this on an actual phone; I'll see if I can remember how to do that).

4) On Firefox, the dropdowns open and close with a mouseover. In chrome, they open with a click, which is fine.  They close if I click outside the box, or on the other dropdown.  I clicked on the arrow in the dropdown a bunch of times thinking "why isn't this closing?!" before it occurred to me that maybe they weren't supposed to.  Should we have the drop-down close on click also? 

@daniellex0 - Most of these are actually design questions. Can you weigh in?  Thanks.

daniellex0:
Looks good!

@drubgrubby - 

-2. Yes, I believe this is because some of the items within the homepage dropdown are longer than the top box so they need more room. It would look cleaner if it was the same size, but this does leave some room for longer team names in the future.. I think for the sake of standardization we can leave it like this for now instead of creating a new class with a different size right now, since this works and looks good enough- but maybe down the line I can add a few different sizes of dropdown options to the design system.

-4. The dropdowns open and close for me when I scroll over on Chrome!

Btw just a small type-o  @akibrhast - (under the div with classname .wins-page-contain projects-inner) - it says id="filter-template-**conatainer**

drubgrubby:
@daniellex0 - What do you think about question 1, where half of the team drop-down is cut off on mobile.  What a good fix for that?

drubgrubby:
@akibrhast @jbubar - Just a general question for this, and all of the other pages that we work on.  Is there any javaScript on this page that might be useful on other pages that we should put in a file to be included, so it can be reused rather than having to be re-written on every page we use it on?  For example, since the drop-downs are basically the same on this page and on the homepage, is there any of that code that is re-useable?

jbubar (approved):
This is really really awesome! I really enjoyed review this code. I learned a lot of new methods and functions! thanks Akib! It works great. One thing is that, as a project, we do not use inline styling. But I think that this can be separated out into a medium sized issue. We will talk about it in the dev meeting today

akibrhast:
> Just a general question for this, and all of the other pages that we work on. Is there any javaScript on this page that might be useful on other pages that we should put in a file to be included, so it can be reused rather than having to be re-written on every page we use it on? For example, since the drop-downs are basically the same on this page and on the homepage, is there any of that code that is re-useable?

@jbubar @drubgrubby The filters that are currently being used in wins-page.html is very similar in what it is doing when comparing to the filter implemented in project-filter.js 

The comment for the class defined in project-filter.js states
```
/**
 * Wraps a Hack for LA project object for filtering
 * @param {Object} data - the object containing all filterable fields for a project 
 */
```
As such it should be possible to create a singular class that can handle that, or maybe subclass from the existing function. Maybe make a new issue/discussion about `Refactoring Filter Methods` . If I am not mistaken #959 could also use this generic functionality,

daniellex0:
> @daniellex0 - What do you think about question 1, where half of the team drop-down is cut off on mobile. What a good fix for that?

@drubgrubby I think it should be formatted the exact same way the dropdowns on the homepage are in mobile if possible - 

Screen Shot 2021-02-14 at 4 55 56 PM

akibrhast:
Made changes as requested by @drubgrubby 
> At a certain point, on mobile, the "Team" dropdown floats off the screen


issue#1030ReworkedFilterCSS

_sass/components/_wins-page.scss
```diff
@@ -274,11 +274,18 @@
 
 //
 ul.filter-list{
```
jbubar: you should not change filter-list because it changes the styling on the main page as well... Use a different class
Screen Shot 2021-02-14 at 5 41 00 PM

_sass/components/_wins-page.scss
```diff
@@ -274,11 +274,18 @@
...
 }
 
-.dropdown{
+ul.filter-list li ul {
 	width: 100%;
-}
\ No newline at end of file
+}
+
+@media (max-width: 659px){
```
jbubar: Maybe we want the breakpoint to happen earlier....
how about using:
@media #{$bp-below-tablet}{
}

jbubar (changes requested):
Two small changes. use new classes, and change the breakpoint

_sass/components/_wins-page.scss
akibrhast: Applied the break point as mentioned above. Removed console.logs

_sass/components/_wins-page.scss
akibrhast: Replaced the css from
``` ul.filter-list{ } ``` which broke the filter styles on the project page to ```.wins-page-contain > ul.filter-list{ } ``` which now only overrides the properties of the filter-list that are a child of the element with class `.wins-page-contain`

jbubar (approved):
Looks good to me!

jbubar:
A good medium issue will be to change the background color to pink when something is checked
Screen Shot 2021-02-14 at 5 51 41 PM

drubgrubby (approved):

drubgrubby (approved):
@akibrhast - Really good work!  This is ready to merge.

PR #1073, opened 2021-02-18
About-us:  fix accordion bug on resize

### LINKED ISSUE #966: About-us: Fix accodion bug on resize
### Overview
When the page is loaded in mobile view, the accordion works.  However, if you then expand the page clicking on the section heading closes the section. 

Screen Shot 2021-01-24 at 7 34 22 PM

Similarly, if you load the page in desktop and then resize the page to mobile, the mobile links don't work. 

### Action Items

Add event listeners that listen for resize or reorientation and reset the page state. 

### Resources/Instructions
https://www.hackforla.org/about-us

drubgrubby:
Fixes #966 

The problem was that the event listeners weren't being created if the page was loaded in desktop, so they weren't available when resized to mobile, and they weren't being destroyed when going from mobile to desktop, so the sections were still opening and closing when in desktop. 

The solution was to create a function to remove the event listeners.  As it turns out, you can only remove an event listener when the event listener calls a function, not when the function is part of the event listener.  So, I had to reorganize the functions to accommodate this. 

The code has an event listener for resize that adds and removes the event listener when the size changes from mobile to desktop.  I created a variable 'accordionFlag' that tracks the state, so you know when it's going from desktop to mobile and only create or remove the event listeners once each time the state changes. 

Finally, there's a function that expands all of the panels when switching to desktop. 

Note: One time I saw an anomaly where teh "Read More" text displayed in the middle of the letter from the executive director.  I could never recreate it, so...  But if you can figure out if it's real, and when it happens, that would be much appreciated.


drubgrubby:
I also removed some comments and console.logs that I was bored with.

drubgrubby:
Just updated a commit to this.  I realized I had a silly double if statement that could have been simpler with some ampersands.  Now it is.

akibrhast:
@drubgrubby Just want to confirm is the read more button on the executive director supposed to be there? Or is that the anomaly you were speaking off?
image

akibrhast (changes requested):
Anomaly of the see more position seems to not be an anomaly on my end.
david

drubgrubby:
@akibrhast - Thanks for helping me figure out when that bug happened.  I believe I have fixed it.  Time will tell what else I messed up in the process.

akibrhast (approved):

akibrhast:
Everything looks good! The anomaly has disappeared! 😄 
Ready for merge

jbubar (approved):
This is really awesome work @drubgrubby! 
I pulled it to my machine, and all my efforts to try to break it were fruitful in finding one bug, Which I think might grow up to be a good issue some day.

if you refresh the page after clicking on one of the nav menu items it ruins the desktop view

Screen Shot 2021-02-18 at 12 17 48 AM

I would like to merge this pr and create a new issue @drubgrubby and @akibrhast do you agree?

drubgrubby:
@jbubar - Thanks for your review.  I'm not sure exactly what you're referring to when you say "ruins the desktop view".  Are you referring to the fact that the menu items no longer track with the sections?  If yes, that is already an issue and Qiqi is working on it currently. 

Either way, I agree that this can be merged, and if the bug is something other than the nav and the sections becoming decoupled, then, yes, make a new issue.

akibrhast:
@jbubar Yeah, should be good for merge. Qiqi is already working on that particular issue that you mentioned.

drubgrubby:
What @akibrhast said.

PR #1072, opened 2021-02-17
Update README.md

ahdithebomb:
Made additional changes per Bonnie's request to update the Readme by linking the Wiki.

PR #1075, opened 2021-02-18
Update README.md

ahdithebomb:
Hi Bonnie,

I had a chance to clean up the verbiage based on our conversation earlier today.


jbubar:
Once Bonnie approves, I believe that we need to merge both pr's that you created, First we merge this one, then we merge #1072... I am pretty sure that if we do it in that order it will avoid git complications.. lol

PR #1078, opened 2021-02-21
Toolkit banner links

mcspach:
-Functionality of category filter seems good. Can someone please review changes.
-I will make the last few CSS changes to banner once the functional part is confirmed.


ExperimentsInHonesty:
Please make sure when you review this page that the text in the body of the cards did not get switched to bold as appears here: https://github.com/hackforla/website/issues/959#issuecomment-782891255. And also that the Titles of the links match the figma.

drubgrubby:
@mcspach - I think this issue isn't quite ready for a PR.  You say in the issue that you still have css changes to make before the final commit.  It's best to wait until you think the page is ready to merge before making a PR.  Also, be sure to put "Fixes" and then the issue number at the top of the PR so that the issue is closed when it is merged. 

I am changing this file to a draft until it's ready for review.

mcspach:
Fixes #959

akibrhast (changes requested):
Ensure that clicking on the filters to not filter out the cards under external links, i.e
- Zoom
- Figma
- GitHub

Figure out why images disappear when filtered.
Ensure that hover state of the filter links should be pink

_sass/components/_toolkit.scss
```diff
@@ -21,13 +21,26 @@
...
         padding: 1.4rem 0;
         background-color: $color-pink;
+        
+        a {
+            font-weight: 300;
+            text-decoration: none;
+            color: black;
+        }
+        a:hover {
+            color: $color-mediumvioletred !important;
```
akibrhast: Do not use !important in your scss files as this is generally bad practice. Could you please remove the !important and achieve the same effect using greater specificity in your css selection.

[Why `!important` is bad practice](https://developer.mozilla.org/en-US/docs/Web/CSS/Specificity#the_!important_exception)

> When an important rule is used on a style declaration, this declaration overrides any other declarations. Although technically !important has nothing to do with specificity, it interacts directly with it. Using !important, however, is bad practice and should be avoided because it makes debugging more difficult by breaking the natural cascading in your stylesheets. When two conflicting declarations with the !important rule are applied to the same element, the declaration with a greater specificity will be applied.

_sass/components/_toolkit.scss
```diff
@@ -21,13 +21,26 @@
...
+            transform: scale(1.01);
+        }
+        .selected-category {
+            color: $color-mediumvioletred !important;
```
akibrhast: Please look at my previous comment about !important

docker-compose.yml
```diff
@@ -3,7 +3,7 @@ services:
   hfla_site:
     image: jekyll/jekyll:pages
```
akibrhast: This file has nothing to do with this pull request. Please remove this file from this pull request -> `docker-compose.yml`

akibrhast (changes requested):
1. The card "Displaying your civic tech" is not rendering its placeholder image when filtered under "Professional Development".
<details>
<summary>Broken Card Render</summary> 

image

</details>

2. The hashmark for both Project Management and Project Development has white spaces in it which is being auto-translated to %20. This is generally bad practice according to url specs. Please ensure that there are no white spaces in the hashmark.

> URL can have an Space Character in them and they will be displayed as %20 in most of the browsers, but browser encoding rules change quite often and we cannot depend on how a browser will display the URL. -  https://stackoverflow.com/a/11176152/10864761

<details>
<summary>Auto encoded white space in hashmark</summary>

image


</details>

_includes/guide-card.html
```diff
@@ -1,6 +1,6 @@
 {% assign guide = include.guide %}
     
-    <div data-article-type={{ guide.category | downcase | remove: ' ' }} class="toolkit-flex-item section-container">
+    <div data-article-type={{ guide.category | downcase | remove: ' ' }} class="toolkit-flex-item section-container guide-card">
```
akibrhast: Try not to add arbitrary class names unless necessary. I understand you added the class name of `guide-card` to ensure that the filters didn't filter out external resources cards which shares the same names. But you already have a specificity selector for this particular <div> tag which is `data-article-type`, making the class name `guide-card` obsolete.

In your JavaScript you could do something like this:
```javascript
let cardContainers = document.querySelectorAll('[data-article-type]');
```

Please remove the `guide-card` class and instead use the data selector for your querySelectorAll

akibrhast (changes requested):

mcspach:
@akibrhast thanks for the thorough review on this one!

Latest push has resolved these issues:
- whitespace issue in url hash has been solved. '%20' has been replaced with '+' to follow site conventions.
Screen Shot 2021-03-08 at 9 25 13 AM

- CSS specificity has been corrected and all !important tags removed.
- Use of additional class for JS query selector has been removed and replaced with data-attribute, unnecessary classname has been removed.
- Unrelated file docker-compose.yml has been removed from pull request.

Issues that are unsolved:
- Broken card render still exists. Some filters do not properly render placeholder image in the card. 
  - I do not have an answer for this problem. 
  - Since this issue has been open a while, I request that this branch be merged and someone assigned to a new issue to fix the image render problem. (hopefully not me because I don't understand the bug). 

 Broken Card Render:
image

akibrhast (changes requested):
@mcspach Great Job on applying the change request. 
When I pulled the application locally, I noticed this weird thing happening with the filter tabs on screen resize. Could please take a look at what's going on?
<details>
<summary>Weird Behaviour On Screen Resize</summary>

pr_1078_hack4la

</details>

In regards to the default images not rendering on filter change. Since you are at a loss on what the issue is. Instead of merging this pull request, I would recommend removing/deleting the pull request from your end.

Then leave a comment on the issue that this pull request is related to , i.e. #959 . On the current status of the your branch, the problems with it, and the required actions on it.

You can then un-assign yourself from the issue, and add a label 'help-wanted' on it.

Make sure that your comment on the issue also states instructions on how to continue working from your branch, For Example:

```bash
// replace with your repo url and branch name and so on
git remote add akibr https://github.com/akibrhast/website
git checkout akibr/wins-works -b wins-works
```

mcspach:
Screen size responsiveness issue is fixed
<details>
<summary>Mobile: NoWrap</summary>
Screen Shot 2021-03-09 at 11 11 11 AM
</details>
<details>
<summary>ExtraLarge: SpacedEvenly</summary>
Screen Shot 2021-03-09 at 11 11 48 AM

</details>

PR #1092, opened 2021-02-22
Implement new GitHub actions for wins page

### LINKED ISSUE #1069: Create a new github action that will eventually replace the wins page github action
### Overview
The current wins data contains all of the information on the [google sheet](https://docs.google.com/spreadsheets/d/1tj6eQlVLFgmskoXouVt87POxYVwR1yWT2vOSQEPyDtg/edit#gid=1106372497) populated by the [google form](https://docs.google.com/forms/d/1RhgDlPikDz63hoIDYQag8pUZwYtKq-ZpGG46mhVOrL0/edit?usp=drive_web)
Because we do not want anything directly published from the site without being currated, we need to create a new github action that only populates the wins data with approved data

### Action Items
- [x] create a script that takes the wins data and converts it into an object with sensible key names that will eventually match the columns on the  new google sheet (#1061) and the new html form (#1060)
- [x] cleans up the wins-data to only include json that has the key value pair 'Display?': 'TRUE', or whatever you decide to name that key
- [x] figure out how to run that script using github actions 
- [x] commit it every night 
- [x] have it generate/update a new file called something like wins-page-data.json so that it does not break the current wins page

### Resources/Instructions
[akib started working on it and has a proof of concept](https://github.com/akibrhast/github-actions-test)

akibrhast:
Addresses #1069 

## Merging this pull request will result in:
1. A new folder called utils , which contains scripts to clean up data pulled from wins form submission before storing it in our repo
2. A new file called  `_data/_wins-data.json`
3. A new GitHub action which creates the file mentioned in 2. under `.github/workflow/_wins-data.yml`

## Action Items
- [ ] Create a new issue before merging this pull request:
  - [ ] It should have a dependency of : Refactor wins.html to work with `_data/_wins-data.json`
  - [ ] It should request the change for the hard coded wins_api_key to be retrieved from secrets instead,
  
## Info
Example json file that is generated can be currently found here
https://github.com/akibrhast/github-actions-test/blob/master/_data/_wins-data.json


cnk:
I am frankly not happy about this using Python for this job. Nothing against Python but this project uses Ruby and JavaScript. I would vastly prefer this use one of those two languages. Would you be up for rewriting it using this as the model? https://github.com/civictechindex/scrape/blob/master/scraper.rb

akibrhast:
Changing this pull request to a draft since there are the things that still needs to be done as specified by @cnk 

- [x] Re-write cleanup script in JavaScript instead of python
- [x] Save new cleanup script under the folder `github-actions`
- [ ] Have it running for 2 days on my fork of hackforla
    - [ ] Have it running on a day when something changes
    - [ ] Have it running on a day when nothing changes

PR #1108, opened 2021-02-24
Fixed 'See our Locations dropdown bug'

Treidexy:
I noticed an issue on the 'Events' page (3 actually), so I fixed one of them, which was that on the bottom, when the width is less than 768px, you need to double-click to see the contents pop out


akibrhast:
Nice Work Teddy!

akibrhast (approved):
Good Catch!

PR #1148, opened 2021-03-02
Remove inline styling

### LINKED ISSUE #1136: Standardize Events Page: Remove all HTML inline styling
### Overview
We need to remove all HTML inline styling from the Events page to keep this page consistent with the rest of the site. (Some has already been removed, but now the rest needs to be done).

### Action Items
- [x] Replace all HTML inline styling with classes and the same SCSS properties

### Resources/Instructions
[Events page](https://www.hackforla.org/)

Example of the inline styling on the page:

Screen Shot 2021-02-28 at 9 44 54 AM

evancchen:
Fixes #1136


qiqicodes (changes requested):
Fixing the inline styling issue caused the h1 of header-container-content.html styling to have #ffffff. And currently it is invisible unless highlighted. 
Screen Shot 2021-03-01 at 10 14 46 PM
Screen Shot 2021-03-01 at 10 15 00 PM

I suggest inline color styling for the h1 of header-container-content.html.

_sass/components/_events.scss
```diff
@@ -523,4 +523,10 @@
 
 .userTimeZone {
   font-weight: 700;
+}
+
+h1 {
+  color: #ffffff; 
+  margin-top: 48px;
```
akibrhast, line 530: Find out from @daniellex0  if this heading styling is a generic one we want to apply to the whole site.
If it is not, make your css selector more specific . For example:
```css
.events-page > h1 {
  color: #ffffff; 
  margin-top: 48px; 
}
```
It would then apply it to only the heading on this page.

However if this heading is a generic one,  Then we would need to discuss further on which file to apply these changes to .

akibrhast (changes requested):

daniellex0:
@akibrhast @qiqicodes @evancchen Sorry I probably should have been more specific in the issue- this definitely shouldn't be a generic h1 class. There should be a specific new class name attached to both "Our Events" and "Our Locations" h1 tags. Something like `.events-subheadings` or whatever you think would be most appropriate. And the properties should be listed under that new class on the `_events.scss` file. Or as Akib said, these h1 properties can be nested under `.events-page`. Whatever you guys think is preferable. But this definitely shouldn't be a new styling that will be connected to all h1 tags across the site and turn them all white!

Also are there supposed to be 14 commits on this pull request? This may relate to checking out back to the gh-pages branch after pushing a branch. We used to have this issue a lot before we added extra emphasis about this in the readme. We can discuss at today's dev meeting..

Issue #752, opened 2020-09-27
Create a Join Us page


ExperimentsInHonesty:
### Overview
We have multiple ways to bring volunteers into hack for LA besides projects, we need a page that provides a pathway for all types of contributions

### Action Items
- [x] Product
   - [x] Add general content to page to be augmented as needed
- [x] UI
   - [x] Mock up an about page that is of the same illustrative design quality as the rest of the site.
- [ ] Development
   - [x]  Create page based on Figma mockup
   - [ ]  Add to menu of HfLA site - desktop and mobile
   - [x] Add screenshots to issue to get visual signoff from product
- [ ] Product
   - [ ] Signoff
- [ ] Development
   - [ ] PR
- [ ] Team review

### Resources/Instructions
[Figma](https://www.figma.com/file/0RRPy1Ph7HafI3qOITg0Mr/Hack-for-LA-Website?node-id=2991%3A0)

yuikomajima:
Here's a pass at the Join us page using the icon you suggested. If we can consolidate the partner with us forms into 1 (by adding a question - are you gov't, non-profit etc.) then it could be cleaner without having 3 CTAs. https://www.figma.com/file/0RRPy1Ph7HafI3qOITg0Mr/Hack-for-LA-Website?node-id=3058%3A21055

ExperimentsInHonesty:
The three forms are slightly different.  But I'm will to take a look again at the questions and see if we can't revise things

alexandrastubbs:
@yuikomajima looks great! Could you play with the circles in the donation section so that they 1) take up a bit more space 2) illustrate the quantity we spend toward each thing (i.e. monthly = most spend, annual = middle spend, occasional = least spent)? Also maybe we change their color (green for money?) and the formatting of the section so that it is more consistent with the other modules.

yuikomajima:
made the above changes. To make the format more consistent, it might be better to have the expense breakdown in text to not have 3 images in one module. https://www.figma.com/file/0RRPy1Ph7HafI3qOITg0Mr/Hack-for-LA-Website?node-id=3058%3A21055

yuikomajima:
please see here for mobile view. Please review and update copy! https://www.figma.com/file/0RRPy1Ph7HafI3qOITg0Mr/Hack-for-LA-Website?node-id=3058%3A21055

alexandrastubbs:
@yuikomajima all good, we updated copy and moved some things around. This is ready for development.

trtincher:
[...]

PR #1158, opened 2021-03-03
Nesting bleeding classes into overarching class

### LINKED ISSUE #1154: Add nesting to page card component in _sitemap.scss
### Overview
We need to nest the page-specific properties of the .pg-card-lg class in the site map page so that they don't affect page cards on other pages.

### Action Items
- [x] Nest the scss properties below (bottom of this issue) to make sure they will only apply to the page cards on the site map page

### Resources/Instructions

These properties should be nested (perhaps under `.sitemap-page` ? ) because they are affecting the page cards on the new Join Us page.

```
  @media #{$bp-tablet-up}{
.page-card-lg:nth-child(1n+2){
      padding-left: 0;
  }
```

(*This is my fault because I am supposed tocreate a generic _page-cards.scss file that all pages on the site can pull from, but I haven't done this yet due to a few dependencies. So now the classes are temporarily living in _sitemap.scss and can easily confuse the situation when these reusable classes need to be used on other pages. I will make sure to create the _page-cards.scss file asap, but in the meantime this is the quick fix!)

ItsLaszlo:
Fixes #1154 

Nested requested classes. Hopefully fixes the bleeding happening on issue #752

### DIFF
--- a/_sass/components/_sitemap.scss
+++ b/_sass/components/_sitemap.scss
@@ -114,44 +114,43 @@
       }
     }
   }
-}
-
-/*---------------------------
-| Base Page Card
----------------------------*/
-.page-card {
-  border-radius: 20px;
-  overflow: hidden;
-  background: $color-white;
-}
-
-/*---------------------------
-| Card Color
----------------------------*/
-// White
-.card-primary {
-  background: $color-white;
-}
-
-/*---------------------------
-| Page Card Size
----------------------------*/
-//Large
-.page-card-lg {
-  max-width: 896px;
-  box-shadow: 0 0 8px 0 rgba(51, 51, 51, 0.2);
-  padding: 48px;
-  margin: 16px auto;
-  // NEW MOBILE FRIENDLY RULES
-  @media #{$bp-below-tablet} {
-    padding: 32px 16px;
-    margin: 12px auto;
+  /*---------------------------
+  | Base Page Card
+  ---------------------------*/
+  .page-card {
+    border-radius: 20px;
+    overflow: hidden;
+    background: $color-white;
   }
-}
 
-@media #{$bp-tablet-up} {
-  .page-card-lg:nth-child(1n + 2) {
-    padding-left: 0;
+  /*---------------------------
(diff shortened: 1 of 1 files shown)


daniellex0:
Hey @martham0 (FYI it's definitely possible that I am looking at the wrong file and you have done this already, or I am confusing this in some way, so take what I say with a grain of salt lol)-

But is it possible to move this section in the purple rectangle-

Screen Shot 2021-03-06 at 6 49 07 PM

... and nest it under this `.sitemap-page` class:

Screen Shot 2021-03-06 at 6 49 15 PM

 ... so that it only affects your page cards and not the ones on Travis' page?

_sass/components/_sitemap.scss
```diff
@@ -40,92 +39,83 @@
...
+      a {
+        display: flex;
+        justify-content: center;
+      }
     }
   }
 }
 
-
 /*---------------------------
```
trtincher: So everything below this still is not nested, if I am looking at this correctly. All of this should be nested as well.

_sass/components/_sitemap.scss
```diff
@@ -149,21 +139,20 @@
...
-      padding: 32px 16px;
-      margin: 12px auto;
-  }
+    padding: 32px 16px;
+    margin: 12px auto;
   }
-
-
-  @media #{$bp-tablet-up}{
-    .page-card-lg:nth-child(1n+2){
```
trtincher, line 164: As this is currently there will still be a bleeding problem I believe.

trtincher (changes requested):

_sass/components/_sitemap.scss
ItsLaszlo: Fixed!

daniellex0 (approved):
I think this should work... but @trtincher you're the nesting expert - if you think this looks good, maybe we should merge this and see how it affects your page?

trtincher (approved):
@martham0 Yeah that will do it I think. Thanks!

PR #1175, opened 2021-03-08
Test for weird behaviour on the wins-work pr

akibrhast:
This pr opens up a url on the live site under hackforla.org/wins-test/
Which exposes the winsTest directory which is a duplicate of the wins directory.

This is being done to try and replicate a weird behavoiour that was seen previously on the live page
when the pr 'wins-work' was merged.


drubgrubby:
The new directory doesn't work, even on my local drive.  Perhaps we can figure this out without having to merge this.  Checking....

PR #1199, opened 2021-03-12
959 toolkit banner links

### LINKED ISSUE #959: Make the Toolkit page's category-filtering banner functional

akibrhast:
Closes #959 

- The majority of the features requested(filtering mechanism) in #959 was implemented by @mcspach 
- I just fixed the last issue on this where filtering was causing images to not render


drubgrubby (approved):
@mcspach @akibrhast - Nice work on these links/filters.  All seems good to me.

jbubar:
Looks great... I just noticed one inconsistency.
If you click on "All", "Development", or "Design" it has a slightly different behavior than clicking  on "Project Management", and "Project Development"

toolkit.html
```diff
@@ -15,73 +15,63 @@ <h1>Our Toolkit</h1>
 </div>
 
 <div class="toolkit-header__text-group">
-    <p class="toolkit-header__banner-item">Development</p>
-    <p class="toolkit-header__banner-item">Design</p>
-    <p class="toolkit-header__banner-item">Project Management</p>
-    <p class="toolkit-header__banner-item">Professional Development</p>
+    {% assign category_list = "All, Development, Design, Project Management, Professional Development" | split: ", " %}
+    {% for category in category_list %}
+      <a href="#{{category | replace: ' ', '+'}}" id="{{ category }}" class="toolkit-header__banner-item">{{ category }}</a>
```
jbubar: This appears to redirect to the guide pages (hiding the menu bar) for All, Development, Design,  but not for Project management, and Professional Development

Notice the href and id are different:

``` HTML
<a href="#{{category | replace: ' ', '+'}}" id="{{ category }}" class="toolkit-header__banner-item">{{ category }}</a>
```

Or.. to be honest. I dont really like the behavior of the <a> tag.. so maybe it would be best to get rid of the href. what do you think?

jbubar (changes requested):
Could you look over this one line? 
line 20 in toolkit.html
It is inconsistent.
Personally, I would prefer if it did not redirect, as in I prefer the behavior of Project Management, and Professional Development rather than All, Development, Design

toolkit.html
akibrhast: > It is inconsistent.
- It is now consistent.

> Personally, I would prefer if it did not redirect, as in I prefer the behavior of Project Management, and Professional Development rather than All, Development, Design 
 - It does not redirect anymore

PR #1200, opened 2021-03-12
Fix buggy scroll

### LINKED ISSUE #1184: Events Page has buggy scroll on mobile
### Overview
The drop downs in the events page has a weird behaviour on mobile(NOT mobile view on desktop). Where scrolling the page fires the resize event which in turn causes the drop downs to auto close while scrolling and trying to read it. The issue is more obvious in firefox browser compared to chrome

### Action Items
- [ ] Prevent drop downs from auto closing when being resized

### Resources/Instructions
- https://stackoverflow.com/questions/17328742/mobile-chrome-fires-resize-event-on-scroll#:~:text=Browsers%20in%20mobile%20devices%20often,client%2C%20firing%20the%20resize%20event.
 - https://gist.github.com/jonnymaceachern/9e2cea7e117f8142a00d

akibrhast:
Fixes #1184 

- Fixes buggy scroll by using greater specificity on the resize event listener

### DIFF
--- a/events/events.html
+++ b/events/events.html
@@ -76,7 +76,7 @@ permalink: /events/
       }
     }
   }
-  window.addEventListener('resize',handleScreenResize)
+  document.querySelector('.flex-page-card').addEventListener('resize',handleScreenResize)
   function handleScreenResize(){
     if(document.body.clientWidth>767){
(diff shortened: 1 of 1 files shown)


jbubar (approved):
Looks good. Clever and simple solution!

PR #1269, opened 2021-03-19
1268 reorganize data only collection to data subdir

### LINKED ISSUE #1268: Proposal to reorganize the data-only collections under a data subdirectory
### Overview
The current state of the repo stands such that any arbitrary group of data that needs to be in a particular page or many pages is made into a collections. This does not conform well to [Jekyll documentations ](https://jekyllrb.com/docs/datafiles/) which states that
>Jekyll supports loading data from YAML, JSON, CSV, and TSV files located in the _data directory. Note that CSV and TSV files must contain a header row.
This powerful feature allows you to avoid repetition in your templates and to set site specific options without changing _config.yml.

The difference between a `collections` and `_data` is that
- _Collection_ — Will have its own page, URLs, and possibly an index page.
- _Data_ — Will not have its own page, but will be referenced to on other pages within the site.
<sub>Further information can be found [here](https://gist.github.com/akibrhast/31c7bb723d80e9efe3d28f619ef0798d)</sub>

### Action Items
- [x] Create a new directory `_data/internal`
- [x] Remove toolkitResources as a collection from the `_config.yml`
- [x] Move the content of toolkitResources to `_data/internal/toolkits/resources.yml`
- [x] Remove press as a collection from `_config.yml`
- [x] Add content of press to `_data/internal/press/press.yml`
- [x] Remove spronsors as a collection from `_config.yml`
- [x] Add content of `_sponsors` to `_data/internal/sponsors/sponsors.yml`
- [x] Remove `_testemonials` as a collection from `_config.yml` and Add content of `_testemonials` to `_data/internal/testimonials/testimonials.yml`
- [x] Make sure nothing breaks or changes

### Resources/Instructions
REPLACE THIS TEXT - Provide links to resources or instructions that may help with this issue. This can include files to be worked on, external sites with solutions, documentation, etc.

akibrhast:
Closes #1268

 - Commit message contains all the changes applied in this PR


akibrhast:
> Two questions:
> 
> 1. Why does _data/internal/events.yml have the --- used to denote front matter while the other yml files in that directory do not? Since this is data, I suspect we probably want to remove the --- from events.yml.
> 2. Why did you choose not to move the communities, credits, and press collections into _data? I think the data in those directories will be easier to manage if we do keep those collections as directories of separate files, but they still fall under 'data' since those files do not in and of themselves create pages.

1. Typo .. I have fixed it now.
2. `since those files do not in and of themselves create pages.` . You are right. I did think about moving them to the data directory. But I was not sure if I should. So ... Should I? @cnk

cnk:
The files currently in this PR are looking good. We can merge this now and decide about the communities, credits, and press collections as another issue if you like. I would like the to be in _data if you want all the non-page thing in _data/internal BUT I think they will be easier to work with as separate files, not one big .yml for each

akibrhast:
> BUT I think they will be easier to work with as separate files, not one big .yml for each

I agree... I just didn't know the syntax to iterate over separate files in the data directory of the top of my head.

> We can merge this now and decide about the communities, credits, and press collections as another issue if you like.

Sounds good
@cnk

jbubar (approved):
Changes look good. Thanks akib

PR #1260, opened 2021-03-18
New pages directory and repo clean up

akibrhast:
Please refer to the commit message in regards to what has been changes

This pull request is a  to test the feasibility of the changes asked in https://github.com/hackforla/website/issues/1138

It also tries to do some clean up work
Please pull this branch for test purposes and try hardest to break it! If possible this should be merged before we start getting in more issues and prs

Closes #1255

PR #1308, opened 2021-03-23
Recreate missing md files back to repo

qiqicodes:
Added following markdown files to repo

* calendar.md
* conduct.md
* github.md
* meetup.md
* slack.md


akibrhast:
@ExperimentsInHonesty 
I have a question about the broken links. It seems they were using a redirection layout.... If I may ask.. why are we not just doing `<a href='link' target="_blank">text</a> ` for those files, like we are doing for other external links? There are a few other places where we are redirecting people but not using the redirection layout. So I was just asking why we are using it in some places and not other places.. ? I have asked both @cnk and @drubgrubby  and they don't know/aren't sure.

jbubar (approved):
Looks like we lost these files in pr #1260.

@akibrhast that is a good question. maybe there is a benefit to having the urls.. hackforla.org/<external>

PR #1318, opened 2021-03-24
Update add cursor pointer to _project-filter.scss

### LINKED ISSUE #1155: Change cursor on drop down of Win page
### Overview
When using the dropdown filters the cursor change isn't accurate. It changes to ``cursor: pointer`` when hovering over anywhere in the dropdown except for the checkbox. On the checkbox, it turns back into a normal cursor. This can be a bit confusing for the user and implies they can select the category by clicking it rather than clicking the checkbox which isn't the case.

https://www.hackforla.org/wins
Screen Shot 2021-03-02 at 6 43 59 PM
Screen Shot 2021-03-02 at 6 44 07 PM

### Action Items

- [x] Have the cursor change to a pointer when hovering over the checkboxes
- [ ] Have the cursor revert back to normal when hovering over the rest of the drop-down

### Resources/Instructions

https://developer.mozilla.org/en-US/docs/Web/CSS/cursor

trtincher:
fixes #1155 
Cursor remains as pointer over checkboxes

https://user-images.githubusercontent.com/35665684/112235916-922fc600-8c0d-11eb-81d5-aa2347a7ca6a.mov

### DIFF
--- a/_sass/components/_project-filter.scss
+++ b/_sass/components/_project-filter.scss
@@ -82,6 +82,7 @@ ul.filter-list li ul li {
   border: 1px solid #000000;
   border-radius: 4px;
   margin: 0px 5px;
+  cursor: pointer;
 }
 
 .labelArrow {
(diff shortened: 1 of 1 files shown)


averdin2 (approved):
Josh thinks this will work great!

PR #1328, opened 2021-03-26
Reworked all redirections into a single collection that all uses the …

akibrhast:
Reworked all redirections into a single collection that all uses the layout redirect.html


jbubar (approved):
This looks good. I like how it keeps things more organized. all of the redirects still work perfectly

Issue #1142, opened 2021-02-28
Consolidate all of the scss files currently attached to the home page into the _home.scss file


daniellex0:
### Overview
We need to combine all of the contents of the scss files that are currently connected to the homepage into the _home.scss file to keep this consistent with other pages on the site (one scss file per page).

### Action Items
- [ ] Find all of the scss files connected to the homepage
- [ ] Move all of the content from those scss files into the combined _home.scss file (created by @kevinreber )

### Resources/Instructions
[Home page](https://www.hackforla.org/)
Example of one of the scss files connected to the home page - [_contact-us.scss](https://github.com/hackforla/website/blob/gh-pages/_sass/components/_contact-us.scss)

Connected to #864

kceddinger:
@sanchece Left instructions on pr #1361

PR #1361, opened 2021-04-06
combined scss files try2

sanchece:
#1142

Moved scss content into _home.scss file
Attempt 2


erikaBell:
I didn't see the `_home-getting-started.scss` put into `_home.scss`. 

Second pair of eyes on this issue would be helpful — @jbubar

PR #1370, opened 2021-04-07
Moved credits collections to credits as data

akibrhast:
The following PR is in line with addressing the comment mentioned here https://github.com/hackforla/website/pull/1269#issuecomment-803459170

> The files currently in this PR are looking good. We can merge this now and decide about the communities, **credits**, and press collections as another issue if you like. I would like the to be in _data if you want all the non-page thing in _data/internal BUT I think they will be easier to work with as separate files, not one big .yml for each - @cnk


drubgrubby:
Looks good.

PR #1373, opened 2021-04-08
Update lucky-parking.md

ExperimentsInHonesty:
change https to http on the demo site so that it will work correctly.


akibrhast (approved):
LGTM. Reason for this change is because http causes the lucky parking site functionality to break

PR #1401, opened 2021-04-13
Events Load Time, Issue 1284

### LINKED ISSUE #1284: Events Page Load Time Slow
### Overview
The events in the events page takes more than half a second to load on every single page refresh

### Action Items
- Cache api data to local storage, and ensure that dont call api on every page refresh unless needed

### Resources/Instructions
- `pages\events\right-col-content.html`
- Store the filtered and sorted object as a string in local storage. Also store the date/and time of the storage as a separate-key
    - if date-time key in storage(means object also in storage) retrieve date-time . Then call api, and see check if any `updateDate:` from api [NEWER THAN] date-time in local storage then do a refilter resort of the object
    - if any `updateDate:` from api is not newer than date-time in local storage.. then dont change dom

erikaBell:
Fixes #1284 

@Akib, I got everything working when testing with "fake updates". The setup for the pseudo API was a bit different than what I had to do to refactor it for our API. So would appreciate a thorough look/check that it is still setup to work with our own.

### DIFF
--- a/pages/events/right-col-content.html
+++ b/pages/events/right-col-content.html
@@ -39,7 +39,6 @@
     (async function main(){
 
         const response_data = await sendGetRequest(API_ENDPOINT);
-
         let filteredData = filterDataFromApi(response_data);
 
         let sortedData = sortData(filteredData);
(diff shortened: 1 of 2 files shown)


pages/events/right-col-content.html
```diff
@@ -31,27 +31,76 @@ <h3 class="day-header-1">{{day}}</h3>
...
+    /**
+     * Stores data and users date-time locally
+     * @param {Object} sortedData — 2 key : value pairs in localStorage
+    */
+     function storeData(sortedData){
+        const dataString = JSON.stringify(sortedData);
+        const userDate = new Date().toISOString(Date()); 
+        
+        localStorage.setItem('storedDate', `${userDate}`)
+        localStorage.setItem('storedData', `${dataString}`)
```
akibrhast, line 78: What is the reason behind two setItems to the same key back to back ? Is the second one not overriding the first one?
```javascript
        localStorage.setItem('storedDate', `${userDate}`)
        localStorage.setItem('storedData', `${dataString}`)
```

erikaBell: They do look very similar but if you look at the last letter on each `key`, you'll see one has an `e` and the other an `a` 

`for datE and datA` (-;

akibrhast (commented):
lgtm

PR #1435, opened 2021-04-22
Changed text in Step 6 content area to match text in Step 6 Table of Contents

### LINKED ISSUE #1434: Fix Step 6 text in section in page of CONTRIBUTING.md to match Step 6 Table of Contents
### Overview
Step 6, "Switch to new issue branch.....", in the Table of Contents isn't jumping to corresponding section in document.

### Action Items
Change text in Step 6 area of document to match text in Step 6 Table of Contents so that it will "jump" to the corresponding section.

### Resources/Instructions
https://github.com/hackforla/website/blob/gh-pages/CONTRIBUTING.md#step-6-change-to-a-new-branch

RobertaFricker:
Fixes #1434 
Changed text so that it's the same in Step 6 Table of Contents link and in the corresponding area of the document.

Per screenshots below:
image

image

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -16,7 +16,7 @@ To develop the site, you'll need to first clone the repository on to your comput
 
 5. [Read Hack for LA's Site Architecture to get acquainted with how the website is structured](https://github.com/hackforla/website/wiki/Hack-for-LA's-Site-Architecture)
 
-6. [Switch to new issue branch before you start making changes](#step-6-change-to-a-new-branch)
+6. [Switch to new issue branch before you start making changes](#step-6-work-on-an-issue-using-git)
 
 
 **After you've worked on your issue and before you make a pull request:**
@@ -181,7 +181,7 @@ docker-compose up
 
 #### Step 5: Read [Hack for LA's Site Architecture](https://github.com/hackforla/website/wiki/Hack-for-LA's-Site-Architecture) to get acquainted with how the website is structured
 
-#### Step 6: Work on an issue using git
+#### Step 6:  Work on an issue using git  
 
 Create a new branch for each issue you work on. Doing all your work on topic branches leaves your repository's main branch (named `gh-pages`) unmodified and greatly simplifies keeping your fork in sync with the main project.
 


CONTRIBUTING.md
```diff
@@ -181,7 +181,7 @@ docker-compose up
 
 #### Step 5: Read [Hack for LA's Site Architecture](https://github.com/hackforla/website/wiki/Hack-for-LA's-Site-Architecture) to get acquainted with how the website is structured
 
-#### Step 6: Work on an issue using git
+#### Step 6:  Work on an issue using git
```
akibrhast, line 184: Can you please ensure that this is a link and not just a clear text.

akibrhast (changes requested):

CONTRIBUTING.md
RobertaFricker: @akibrhast I agree that it's not a true hyperlink.  It's actually referred to as an "anchor jump link."

Per https://coder-coder.com/make-html-anchor-link-jump-specific-part-of-page/, to create this, "you need a link and a target. The target can be any element on the web page that has an “id” attribute."
Also, "If we were on a page and wanted to jump down to a section on the same page, we could create an internal link. An internal link doesn’t require the https://www.WebsiteName.com in the link’s href attribute."

For CONTRIBUTING.md, the internal jump link is: 

image

and the target has the ID per the screenshot below:

image

The issue is the jump link does not have the same text as the ID text.

akibrhast: > I agree that it's not a true hyperlink


Ah, my bad. I should have clarified... Can you please ensure that it is a true hyperlink. Please refer to the syntax used in step 5.

RobertaFricker: @akibrhast ...lol...href="#step-6-work-on-an-issue-using-git" is not a true hyperlink, it doesn't go anywhere but within the same page to the area that matches its corresponding target.  It doesn't open a new tab and it doesn't have https:// in its syntax.

akibrhast: My apologies. I was a bit confused initially. Everything looks good!

akibrhast (approved):
lgtm

RobertaFricker:
@akibrhast   I just checked and it still isn't working because there is an extra dash in the target:

image

And, I'm not finding that code block in VS Code, it looks like the screenshot below:

image


Please advise.

PR #1484, opened 2021-04-28
Adjust toolkit page

### LINKED ISSUE #1463: Adjust the Toolkit page filter banner
### Overview
We need to adjust the Toolkit page filter banner so that it functions as intended, and so that its colors are consistent with our design system

### Action Items
- [ ] Change the hover text pink color to our standard HfLA pink (using the variable)
- [ ] Change the black text color from 'black' to our standard dark text variable
- [ ] Make the selected filter bold (font-weight: 700)
- [ ] Make the default selected filter All - so when you arrive at the page, All should be pink and bolded

### Resources/Instructions
[Toolkit page](https://www.hackforla.org/toolkit/)

Toolkit filter banner for reference:

Screen Shot 2021-04-25 at 9 09 59 AM

evancchen:
Fixes #1463


_sass/components/_toolkit.scss
```diff
@@ -26,15 +26,16 @@
...
             a:hover {
-                color: $color-mediumvioletred;
+                color: $color-red;
                 transform: scale(1.01);
             }
 
             .selected-category {
-            color: $color-mediumvioletred;
+                font-weight: 700;
+                color: $color-mediumvioletred;
```
erikaBell: Overall, changes look good! Just want to confirm with @daniellex0  if this is supposed to be _pink_ instead of _mediumvioletred_?

Basing this question of off Danielle's comment in the issue:

- "Make the default selected filter All - so when you arrive at the page, All should be pink and bolded"

My one other comment @evancchen, would be to please show before and after photos so it's easier for the reviewer to know what changes to look for on the local site. Thanks ~

daniellex0: Yes, good catch @erikaBell - @evancchen please change $color-mediumvioletred to $color-red here (I know it's called red but it's actually pink, it's confusing..!)

And agreed, screenshots would be great next time!

daniellex0:
This looks great Evan! Just make the variable change @erikaBell mentioned above and this should be good to go 👍

erikaBell (changes requested):
Change color of `.selected-category` — use comments above for reference.

erikaBell (approved):
Thanks Evan. looks good.

PR #1624, opened 2021-05-24
Standardize About page header and body in mobile

### LINKED ISSUE #1605: Standardize About page header and body in mobile
### Overview
We need to standardize the About page header and body in mobile to make it consistent with other pages and our design system.

### Action Items
- [ ] Decrease the About page header paragraph text font size to 16px **(only in mobile)**
- [ ] Change margin between "About" heading and header paragraph to **28px** (instead of 30px) **(only in mobile)**
- [ ] Increase the margin between the header paragraph and header illustration to 36px total (only in mobile)
- [ ] Figure out why ".page-content-container-grid" class is adding extra margin below the header illustration and a grid-gap to the right of all of the page cards on the About page in mobile, pushing them all to the left
   - [ ] Remove this without affecting the desktop version of the page - only mobile

### Resources/Instructions
[About page](https://www.hackforla.org/about/#letter)

.page-content-container-grid class causing extra margin and grid gap:

Screen Shot 2021-05-22 at 7 04 36 PM

erikaBell:
Fixes #1605 

I put the "About Us" header to 16px/1rem as suggested, but let me know?

<details>
<summary>Before & After: 1</summary>

Screen Shot 2021-05-23 at 7 22 09 PM

Screen Shot 2021-05-23 at 7 22 16 PM

</details>

<details>
<summary>Before & After: 2</summary>

Screen Shot 2021-05-23 at 7 22 26 PM

Screen Shot 2021-05-23 at 7 22 41 PM

</details>

### DIFF
--- a/_sass/components/_about.scss
+++ b/_sass/components/_about.scss
@@ -942,3 +942,23 @@ a.anchor {
     margin-left:3vw;
   }
 }
+
+@media #{$bp-below-tablet} {
+
+  .about-header-text-margin h1 {
+    font-size: 1rem; // equal to 16px
+    margin-bottom: 28px;
+  }
+  
+  .header-text--about-us {
+    margin-bottom: 36px;
+  }
+
+  .page-content-container-grid {
+    grid-template-columns: none; 
+  }
+  .sticky-nav-content-container {
+    margin-left: 0px;
+  }
+
+}
--- a/_sass/components/_about.scss
+++ b/_sass/components/_about.scss
@@ -600,6 +600,20 @@ a.anchor {
 
 /* Below are the breakpoints and adjustments for different size monitors */
 
+// Below tablet - 768px
+@media #{$bp-below-tablet} {
+  .about-header-text-margin h1 {
+    margin-bottom: 28px;
+  }
+  .header-text--about-us {
+    font-size: 16px;
+    margin-bottom: 36px;
+  }
+  .page-content-container-grid {
+    margin-top: 0; 
+  }
+} // End below tablet
+
 // Bigger than mobile - 480px
 @media #{$bp-mobile-up} {
 
@@ -854,6 +868,7 @@ a.anchor {
 } // End desktop-up
 
 // These are to control the responsive location of the sticky nav
+
 @media only screen and (max-width: 1321px) {
   .page-content-container-grid {
     grid-template-columns: auto auto;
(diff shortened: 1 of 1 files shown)


averdin2:
@daniellex0 The issue specifies that the text font size for the About header be 16px, but this seems a bit small to me. The Figma file has it at 24px and that looks better to me. Should it be 24px instead of 16px? Also, @erikaBell what do you think?

daniellex0:
@averdin2 Yes I'm so sorry, I wrote that in a really confusing way- I meant that the paragraph text in the header should be 16px, not the heading- sorry about that!! The "About" heading text should stay 32px. I should have included an image of how it should look for reference, my bad!! 

I also just updated the heading to 32px from 24px on the About page Figma file- sometimes on Figma we are behind the standardization on the design system too unfortunately, but trying to update everything..

_sass/components/_about.scss
```diff
@@ -942,3 +942,23 @@ a.anchor {
     margin-left:3vw;
   }
 }
+
+@media #{$bp-below-tablet} {
+
+  .about-header-text-margin h1 {
+    font-size: 1rem; // equal to 16px
```
daniellex0: 2rem instead of 1rem, sorry about the confusion

daniellex0 (changes requested):

averdin2 (changes requested):
@erikaBell Good job on this issue. I found that around lines 604 to 614, this seems like the best place (code readability wise) in the _about.scss file to actually place the code you wrote. 

I also noticed that the first part of the fourth action item on the issue to "Figure out why '.page-content-container-grid' class is adding extra margin below the header illustration" wasn't completed. I think removing margin-top: 20px on line 51 might be the solution.

erikaBell:
Ah, good catch on the header margin @averdin2, had missed that when reading through the fixes. 

I went ahead and moved some of the code up to where you suggested, the rest of it I had to keep below in order for it to work, due to the hierarchy of the CSS structure.

I believe everything should be good now? Lmk!

daniellex0 (changes requested):
This looks really great, thank you so much @erikaBell !! And thanks for the catches as always @averdin2 !

Ok so the only small change is that when margin-bottom: 36px is added, it looks like it adds the 36px margin below each paragraph- but it's only supposed to be 36px below the bottom paragraph and above the illustration (the margin between the two paragraphs should just stay the automatic one that comes with the < p > tags). I think the solution for this might be to remove the "margin-bottom: 36px" from ".header-text--about-us" class, and instead put both of the < p >'s with class "header-text--about-us" in a div with its own "margin-bottom: 36px" property? Is that too many layers? 

Anyhow you can decide however you think is best to solve it.. Let me know if any of this doesn't make sense! And sorry that I left these specifications out- this is just what happens when I get lazy and don't add an annotated image lmaoo

erikaBell:
@daniellex0 good catch ~ I believe it's all in order now (-: again, just lmk

daniellex0 (approved):
I'm so sorry about the delay- this looks perfect, thank you! 🥇

erikaBell:
@averdin2 Can't merge this myself, so reaching out to you.

averdin2 (approved):
@erikaBell These changes look good to me. Great job! I will merge 👍

PR #1636, opened 2021-05-25
Standardize sitemap header class

### LINKED ISSUE #1608: Standardize Sitemap page header class
### Overview
We need to standardize the Site Map page header class to make it consistent with other pages and our design system.

### Action Items
- [ ] Wrap standard ".header-container" class around Sitemap page header (instead of only the current ".header-container--sitemap" class)
   - [ ] leave any additional properties (flex-direction etc.) in that additional ".header-container--sitemap" class
   - [ ] See other pages that utilize these classes as an example- [Credits page](https://www.hackforla.org/credits/), etc.
   - [ ] Make sure the Site Map page still looks the same in mobile and desktop after changing classes

### Resources/Instructions
[Sitemap page](https://www.hackforla.org/sitemap/)

grhhhh:
Fixes #1608 

Adds `header-container` class to sitemap header `<div>` like on other pages (e.g. the Credits page).


arghmatey (approved):

daniellex0 (changes requested):
Thank you for your work on this @grhhhh !!

The standard header-container class is added great, but the only issue (and this is my fault for not clarifying this at all in my issue so no worries) is that some of the properties in the "header-container" and "header-container--sitemap" classes are now overlapping, and so the properties of "header-container--sitemap" are automatically getting prioritized over those of "header-container."

So it looks like the solution will be to remove the following properties from the "header-container--sitemap" class please:

background: # fff
padding: 64px
height: fit-content
margin: 37px auto 0
font-family: 'Roboto' sans-serif
font-style: normal
font-weight: normal
text-align: center

Those properties (or the correct version of them in the case of the padding) from "header-container" should now replace those and everything should still look good. But please double check and make sure everything looks alright still in desktop and mobile.

Thank you!!

daniellex0 (approved):
Looks great, thanks @grhhhh !!

PR #1681, opened 2021-06-02
removed flex property that caused bug with CSS in Firefox

### LINKED ISSUE #1592: Vertically center the Homepage filter tags text
### Overview
We need to vertically center the text on the homepage filter tags so that they match the original Figma design. They used to look correct, but some update apparently changed them recently.

### Action Items
- [x] View the filter tags on the [homepage](https://www.hackforla.org/#) (select any filter item and the pink tag will show up)
- [x] View the original homepage filter tags design in[ Figma](https://www.figma.com/file/0RRPy1Ph7HafI3qOITg0Mr/Hack-for-LA-Website?node-id=352%3A205) (in pink rectangle)
- [ ] Adjust the alignment of the homepage filter tags text so that they are properly vertically centered again

### Resources/Instructions
[Homepage](https://www.hackforla.org/#)
[Homepage Figma](https://www.figma.com/file/0RRPy1Ph7HafI3qOITg0Mr/Hack-for-LA-Website?node-id=352%3A205) (in pink rectangle)

How the filter tag currently looks (incorrect):
Screen Shot 2021-05-19 at 1 23 52 PM

How the filter tags look on Figma (correct):
Screen Shot 2021-05-19 at 1 24 12 PM

jamesberke:
fixes #1592 

flex:auto was not being used in chrome and was overriding the actual CSS of the tabs in firefox because of the flex grow and flex shrink properties.

### DIFF
--- a/_sass/components/_project-filter.scss
+++ b/_sass/components/_project-filter.scss
@@ -103,7 +103,6 @@ ul.filter-list li ul li {
 }
 
 .filter-tag {
-  flex: auto;
   align-self: flex-start;
   height: 24px;
   box-sizing: border-box;


erikaBell (approved):
Nice, lgtm!

PR #1701, opened 2021-06-08
Standardized the page card classes in the About page

### LINKED ISSUE #1553: Standardize the page card classes in the About page
### Overview
We need to standardize the page card classes used in the About page to be consistent with the Design System

### Action Items
- [ ] Read the [wiki about standardized components ](https://github.com/hackforla/website/wiki/Standardized-Components)(reusable classes)
- [ ] Review the About page's current page card class (just ".page-card--about")
- [ ] Try to figure out which classes to use instead for the page cards on the About page from [this reusable card class file](https://github.com/hackforla/website/blob/gh-pages/_sass/elements/_cards.scss) 
   - **leave out** the top page card- keep as-is. It is different on purpose (no corner radius) because it is supposed to look like a letter
   - note that the page card width should be 896px (so ".page-card-lg" for the size)
   - note that the page-specific class can include the specific padding for the About page
   - Ask @daniellex0 any questions (this is a test run to see how devs use the reusable classes without additional documentation, so that we can know which additional information is necessary)
- [ ] Make sure the About page still looks the same
- [ ] Make sure it still looks right in mobile/tablet view
- [ ] Add @daniellex0 as one of the reviewers for the PR

### Resources/Instructions
[About page](https://www.hackforla.org/about/#accomplishments)[Page cards standardized component classes](https://github.com/hackforla/website/blob/gh-pages/_sass/elements/_cards.scss)[Standardized Components wiki](https://github.com/hackforla/website/wiki/Standardized-Components)

Screen Shot 2021-05-16 at 8 45 09 AM

sena-ji:
Fixes #1553

I tried to include most of the standardized classes as much as possible without changing things. I know it was mentioned to not touch the letter from the ED but if I deleted the white background from the .page-card--about class and use .card-primary class, the letter itself did not have a white background. So added the .card-primary to the letter from the ED. Same with the box shadow around the edges, I added it directly to the page-card--ed class itself since we only want the box shadow from the .page-card-lg class.

I wanted to also use the standardized class .page-card-base because we can then remove the overflow: hidden rule from the direct .page-card--about class. However the css rule width: 100% was messing up the mobile view of the page. So I decided to keep the overflow: hidden rule in the .page-card--about class.

@daniellex0 let me know if you need me to make other changes or if you need some clarification!

Here is a desktop view after changes are made:
<details>
<summary>Desktop View [Click Here!]</summary>

image

</details>

Here is a mobile view after changes are made:
<details>
<summary>Mobile View [Click Here!]</summary>

image

</details>

### DIFF
--- a/_sass/components/_about.scss
+++ b/_sass/components/_about.scss
@@ -143,12 +143,9 @@ a.anchor {
 }
 
 .page-card--about {
-  background: #fff;
+  overflow: hidden;
   border: 0 solid rgba(51, 51, 51, 0.06);
   border-radius: 16px;
-  box-shadow: 0 0 8px 0 rgba(51, 51, 51, 0.2);
-  overflow: hidden;
-  max-width: 896px;
   margin-bottom: 0;
   padding: 22px;
   height: fit-content;
@@ -157,6 +154,7 @@ a.anchor {
 
 // Make the Letter page card have square corners
 .page-card--ed {
+  box-shadow: 0 0 8px 0 rgba(51, 51, 51, 0.2); 
   border-radius: 0;
 }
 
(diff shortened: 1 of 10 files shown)


akibrhast (approved):
Code side everything looks great! @daniellex0 Do merge this if everything looks good to you in terms of the stylistic changes!

daniellex0 (approved):
@sena-ji ok got it yes this will work, thanks!

PR #1745, opened 2021-06-15
revert wins-data.json

jbubar:
Need to revert the data because it disappeared! No idea why yet. but figuring that out is the next step


akibrhast (approved):
Lgtm

PR #1758, opened 2021-06-16
Update comment on gh action add-issue-labels-to-pr

akibrhast:
The comments added to the action clarifies and notes the decision behind the usage of the event
`pull_request_target` instead of `pull_request`

### DIFF
--- a/.github/workflows/add-issue-labels-to-pr.yml
+++ b/.github/workflows/add-issue-labels-to-pr.yml
@@ -1,6 +1,11 @@
 # Name that appears on the workflow
 name: Add Linked Issue Labels to Pull Request
 on:
+  # github actions triggered by a pull request runs with only read only context due to security and safety reasons(this is the expected behaviour),
+  # as such the event trigger `pull_request` could not be used in this action, since this action needs to make a repository edit.
+  # To bypass this github has provided the event trigger `pull_request_target`, more information regarding this can be found here
+  #       - https://github.blog/changelog/2021-02-19-github-actions-workflows-triggered-by-dependabot-prs-will-run-with-read-only-permissions/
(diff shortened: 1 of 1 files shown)


jbubar:
@akibrhast happy to merge this. just comments. but before I do could you explain why the checks are failing? Pretty sure its not a big deal

Aveline-art:
@jbubar The checks are failing because there is no linked issue. The fact that it failed means it is working as intended!

Also, feel free to merge. I still need to make edits and checks to patch up some potential security concerns and some known bugs. I'll open a new issue for that.

akibrhast:
Yes. To everything @Aveline-art said. Would appreciate the merge @jbubar , thanks!

@Aveline-art I am thinking, "All checks have failed" in bright red highlight is not the behaviour we want because sometimes there does not exist a a linked issue for a PR. Case and Point being this PR .

- Since the purpose of the script is to "Add Linked Issue Labels to Pull Request" I believe the script should just skip and end completely if there are no linked issues at all, instead of returning a failure message.
- Another thought that has come to mind is that the current script cannot mirror the issue labels. i.e if the label on an issue is updated it does not translate to a PR label update. This behaviour is probably only limited to 'status' type labels, I believe because generally labels tend to be stable. So maybe, the status type label should not mirrored at all in the first place ?

Aveline-art:
@akibrhast I have not looked into it much, but I am assuming that we would need some level of hard-coding to exclude status labels. I am somewhat wary of that, since the code would be harder to maintain.

I actually want to open up a conversation about reorganizing the labels a bit as well, so the labels might change in the near future. How much of a user issue is mirroring the status labels, as of right now?

Edit: New issue to continue discussion: #1787

PR #1990, opened 2021-07-20
Fix CoP page header image size

### LINKED ISSUE #1793: Fix Communities of Practice page header image size
### Overview
We need to fix the Communities of Practice page's header image so that it matches the original Figma design, and looks more proportional with the text

### Action Items
- [x] Figure out why the header image in the [Communities of Practice](https://www.hackforla.org/communities-of-practice) page is currently 300x240 size when it is supposed to be 384x240
- [x] Fix to make sure it now has the correct measurements

### Resources/Instructions
[Communities of Practice page](https://www.hackforla.org/communities-of-practice)
[Hack for LA site architecture ](https://github.com/hackforla/website/wiki/Hack-for-LA's-Site-Architecture)

How header image looks now (incorrect- too small - 300px width)
Screen Shot 2021-06-20 at 12 41 48 AM

How header image is supposed to look (correct- 384px width)- I got this to be correct here in inspect element when I tested changing the font size- suddenly the image turned into the correct size. Then I switched the font size back to its original size, and the image stayed large (the correct size). Not sure what that means...
Screen Shot 2021-06-20 at 12 41 58 AM

macho-catt:
Fixes #1793 

Looks like the issue occurs only on chrome and edge. Firefox did not have the issue. If you just keep on refreshing on chrome or edge, the image would randomly become smaller or the normal size. Firefox would stay the same on each render.

There are two possible solutions for the issue (as mentioned in my comment in 1793). I went with option 2 since no other pages are using object tags for SVGs, and this keeps me from messing with the CSS and other possible media query adjustments.

  - Changed the object tag to img tag for the community of practice header image to fix the sizing issue + be consistent across the site
  - Changed the object tag to img tag for the slack icon to be consistent with the above + rest of the site

<details>
<summary>CoP header image before</summary>

before-fix

</details>

<details>
<summary>CoP header image after</summary>
  
after-fix

</details>

<details>
<summary> Image responsiveness after fix</summary>

responsiveness

</details>

<details>
<summary> Full page after fix</summary>

chrome-full-after-fix

</details>

### DIFF
--- a/pages/communities-of-practice.html
+++ b/pages/communities-of-practice.html
@@ -14,11 +14,11 @@ permalink: /communities-of-practice
             If you'd like to join one of the Communities of Practice that has a regular meeting time, join their Slack channel and look for the meeting details, usually pinned or in the description. Otherwise, see the "How to get started" sections below.
         </p>
             <div class="header-self-invite--communities">
-                <object data="/assets/images/communities-of-practice/slack-self-invite.svg" type="image/svg+xml" class="self-invite-img" /></object>
+                <img class="self-invite-img" src="/assets/images/communities-of-practice/slack-self-invite.svg" alt="slack icon" />
                 <p class="alert--communities">To join the Slack channels below, you first need to be a member of our Hack for LA Slack Community – you can self-invite <a href="https://hackforla-slack.herokuapp.com" target="_blank">here</a>.
             </p></div>
     </div>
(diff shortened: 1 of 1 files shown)


erikaBell (approved):
Strange bug. 

Everything lgtm~* Nice work @macho-catt (-:

PR #2060, opened 2021-08-04
GH-Actions: Add missing labels

### LINKED ISSUE #2013: GitHub Actions: Adding missing labels to issues lacking the minimum 3
### Overview
As a developer on the team, I need to know exactly what an issue entails so that I can quickly filter for one that is appropriate for my role and skill level.

### Scope

Our team currently requires every issue to have at minimum 3 labels: a size label, a feature (**or** p-feature) label, and a role label. This GHA will alert the team member, who created the issue, that at least one of the labels is missing.

### Psudo-code
- [x] Before starting, scroll through our labels to get an understanding of what this issue is about.
- [x] Trigger: on issue creation
  - [x] If the issue lacks one of the 3 labels:
    - [x] Add any of the 'Feature: missing', 'role: missing', or 'Size: missing' labels as appropriate.
    - [x] Post a comment on the issue alerting them that one of the three labels are missing.
    - [x] The comment needs to have
      - [x] an @username to the creator of the issue
      - [x] a list of the label types that are missing
      - [x] a reminder to also move their issue into the project board and the new issue creation column
      - [x] a link to GitHub's documentation on how to 1) add labels, 2) move an issue into a column
      - [x] instructions to remove the missing labels
  - [x] Else:
      - [x] Post a nice comment indicating that the assignee has done well with the labels
        - [x] The comment should also include a friendly reminder to also move their issue into the project board and the new issue creation column
- [x] In the interest of maintenance, the comment should exist as an md file outside of the code. This allows changing the comment without diving into the code itself. Any unstable variables should be noted with placeholders. See [here ](https://github.com/hackforla/website/blob/gh-pages/github-actions/pr-instructions/pr-instructions-template.md) for an example.

### Checks

- [x] When an issue is created with all 3 of the minimum labels: no comment is made.
- [x] When an issue is created with 'role: design' and 'size: small' and no other labels: a comment is made noting that a feature or p-feature is missing.
- [x] When an issue is edited with any number of labels (0-positive inf), no comment is made.
- [x] Your pull request contains at least 1 md file.

### Additional items

- [ ] If the status label wiki in #1983 or the issue wiki in #1941 is finished
  - [ ] Tack on links to the above mentioned wikis into the automatic comment made by this GHA.
- [x] Else, if either is not done
  - [x] Tack on additional action items to one or both issue as appropriate, noting that the automatic comment needs to include links to those wiki.

### Resources/Instructions

#### Additional resources:

macho-catt:
Fixes #2013

  - Added gh-action to add the required missing labels on an issue if a user forgets them.
  -  The gh-action also removes a "missing label" if the required label is actually filled out. (i.e. I accidentally added the "role missing" label even though I added the "role: front end" properly. The gh-action will remove the "role missing" label and keep the proper label)
  - The gh-action also posts a comment depending on if the user added the required labels or not

### Screenshots of example issues on my fork

<details>
<summary>Issue with no labels</summary>

Screenshot from 2021-08-03 19-03-39

</details>

<details>
<summary>Issue with required labels</summary>
  
Screenshot from 2021-08-03 19-03-59

</details>

<details>
<summary>Issue with extraneous labels</summary>
  
Screenshot from 2021-08-03 19-04-15

</details>


github-actions/issue-labels/check-labels.js
```diff
@@ -0,0 +1,91 @@
...
+ * @param {Object} c - context object 
+ */
+
+async function main({ g, c }) {
+  github = g
+  context = c
+  const issueNum = context.payload.issue.number
+
+  const labels = obtainLabels()
+  const filteredLabels = await filterLabels(labels)
```
alexeysergeev-cm: Is this function `async` at all? It seems it does not need `await` or am I missing something?

github-actions/issue-labels/post-labels-comment.js
```diff
@@ -0,0 +1,77 @@
...
+    return `Good job @${issueCreator} for adding the required labels to this issue.`
+  }
+
+  return `Hi @${issueCreator}. Please don't forget to add the proper labels to this issue.
+  Currently, the labels for the following are missing:
+  ${labels.map(label => ` ${LABELS_OBJ[label]} `)}
+  To add a label, take a look at Github's documentation [here](https://docs.github.com/en/issues/using-labels-and-milestones-to-track-work/managing-labels#applying-a-label).
+  Also, don't forget to remove the "missing labels" after you add the proper labels.
+  To remove a label, the process is similar to adding a label, but you select a currently added label to remove it.`
+}
```
alexeysergeev-cm: Is there an option we can put this in a separate `.MD` file and just pass the username to avoid touching code if the change in the message is needed? 

Can we also remove spaces in the ` ' ${LABELS_OBJ[label]}  '` on line 49 to avoid the spacing in the comment, or is that necessary?

alexeysergeev-cm (commented):
Hi @macho-catt, amazing work! I like how the code is easy to read!

github-actions/issue-labels/check-labels.js
macho-catt: I can remove the async. Previously it was an async function that I refactored, but I missed removing this part. Good catch.

github-actions/issue-labels/post-labels-comment.js
macho-catt: I can do that. It might actually be better that way. Thanks for the suggestion!

macho-catt:
Updates based on @alexeysergeev-cm suggestions

### Additional changes

  - Changed the filterLabels function under check-labels.js to be synchronous
  - Added new MD files for comment templates, instead of only having one:
    - labels-instructions-template.md - overall comment to be posted
    - add-labels-template.md - instructions template for labels that need to be added
    - no-labels-template.md - comment if required labels are added
    - fixed spacing for the labels in the posted comment
  - Updated logic for creating a comment
  - Updated formatComment function to be more usable outside of this issue
  - Updated postComment function to be more usable outside of this issue
  - Separated the two functions above into the utils folder so other files can use them

### Some comments
Since the format comment and post comment are used in other gh-action js files, (the overall function and result are the same), I extracted them into their own files. I noticed that other js files (such as post-comment.js under pr-instructions and add-label.js under add-update-label-weekly) use them as well. They might need to be re-edited a bit, but I think it'll be good moving forward to use this. What do you think @Aveline-art ? Also, I saw two labels that don't follow the format of other labels in the same group: "Role: Administrative" and "Role: Product". Can these be changed to follow the format "role: xxxx" since that's how the other roles are formatted?

It should also be noted that if any of the size, roles, or features labels are edited / updated in the feature, the new code stemming from this PR should be noted since the code uses the pattern of how the current labels are.

### Screenshots of example issues on my fork

<details>
<summary>Updated issue with no labels</summary>

no labels


</details>

<details>
<summary>Updated issue with required labels</summary>
  
with-labels


</details>

github-actions/utils/comments.js
```diff
@@ -0,0 +1,41 @@
...
+      owner: context.repo.owner,
+      repo: context.repo.repo,
+      issue_number: issueNum,
+      body: comment,
+    })
+  } catch (err) {
+    throw new Error(err);
+  }
+}
+
```
Aveline-art: I am going to veto this, since it also requires a lot of parameters, and the catch function should be customizable according to the preferred performance of the function. API calls by its nature already reduce the needed parameters as much as possible, so I think trying to reduce it even more takes away some of the flexibility of it.

github-actions/utils/comments.js
```diff
@@ -0,0 +1,41 @@
...
+ * @param {String} placeholderString - the placeholder to be replaced in the md file
+ * @param {String} path - the path of the md file to be formatted
+ * @param {String} textToFormat - the text to be formatted. If null, use the md file provided in the path. If provided, format that text
+ * @returns {String} - returns a formatted comment to be posted on github
+ */
+ function formatComment(replacementString, placeholderString, path, textToFormat) {
+  const text = textToFormat === null ? fs.readFileSync(path).toString('utf-8') : textToFormat
+  const commentToPost = text.replace(placeholderString, replacementString)
+  return commentToPost
+}
```
Aveline-art: I can see why this should be here, but I also hesitate since 4 required parameters seem to defeat the purpose of making an outside function. I think a utility function should be easy to understand, and does one task very well. I can also envision the need for multiple placeholders, which this does not account for. Not sure if either of those concerns can be addressed, but if not, just put this back into the main folder..

github-actions/issue-labels/labels-instructions-template.md
```diff
@@ -0,0 +1,13 @@
...
+Hi @${issueCreator}.
+
+${labelInstructions}
+
+Just a friendly reminder to also move the issue into the "New Issue Approval" column under the Project Board. For more info on that, check Github's documentation [here](https://docs.github.com/en/issues/organizing-your-work-with-project-boards/tracking-work-with-project-boards/adding-issues-and-pull-requests-to-a-project-board).
+Thanks!
+
+**Additional Resources:**
+Wiki: How to add status labels to issues (_WIP. Link will be updated when the wiki is done_)
+Wiki: How to create issues (_WIP. Link will be updated when the wiki is done_)
```
Aveline-art: I think we would prefer to have two files, rather than a template file that can be adjusted with another file. The idea behind the outside md file is so that a user can clearly see what message the client is seeing (or in other words, a PM, with no programming experience, can adjust their message and get a clear idea of what their message is). Can this be moved into the other templates? It is okay if the files are very similar; it is much more important that we get a grasp of what message we are sending.

github-actions/issue-labels/check-labels.js
```diff
@@ -0,0 +1,106 @@
...
+ * @param {Array} labels - array of labels to check
+ * @return {Array} - returns an array of the labels to add
+ */
+function checkLabels(labels) {
+  let labelsToAdd = []
+
+  REQUIRED_LABELS.forEach((requiredLabel, i) => {
+    const regExp = new RegExp(`\\b${requiredLabel}\\b`, 'g')
+    const isLabelPresent = labels.some(label => regExp.test(label))
+
```
Aveline-art: I've noticed an issue with the role vs Role labels. I think the [i modifier ](https://www.w3schools.com/jsref/jsref_regexp_i.asp)would help with that? I am not sure if there could be another edge case, but if so, let me know so I can make adjustments.

github-actions/issue-labels/check-labels.js
```diff
@@ -0,0 +1,106 @@
...
+  const issueNum = context.payload.issue.number
+  const owner = context.payload.repository.owner.login
+  const repo = context.payload.repository.name
+
+  // Use a flattened set to ensure each label is unique
+  const labels = [...new Set([
+    ...labelsToAdd,
+    ...currentLabels
+  ])]
+
```
Aveline-art, line 89: Not sure why the set object needs to be unpacked. Does the API not take a set object?

Aveline-art (changes requested):
Great work on this! Just a couple of aesthetic changes for this one. Once those are addressed, I can start testing these changes.

github-actions/utils/comments.js
macho-catt: Okay, I can move it back to the original folder

github-actions/utils/comments.js
macho-catt: I'll also move this back as well.

github-actions/issue-labels/labels-instructions-template.md
macho-catt: I'll consolidate this and add it into the two files

github-actions/issue-labels/check-labels.js
macho-catt: I could try that, that might actually make it easier

github-actions/issue-labels/check-labels.js
macho-catt: I only wanted to use the Set object to clear duplicates when I merge the two arrays, and initially when I wrote this and passed it on I wanted to retain it as an array so I can use any properties that come with it (for example, I can use map with an array but with a set I have to use the spread syntax first to transform it to an array). I also wanted to remove any unforeseen effects in using a Set object, so to make my life easier I made it an array.

I could change it back to just a Set if there's a need for that, but otherwise I think using it as an array will make it easier to use.

macho-catt:
### Additional changes

  - Removed the utils/comment.js file and added the two functions there back to post-labels-comment.js
  - Removed the labels-instructions-template.md file and added the body from that into the two other md files
  - Changed the input parameters for formatComment to take an object instead of four parameters.
    - Pros and Cons of using multiple parameters vs an options object: https://stackoverflow.com/questions/12826977/multiple-arguments-vs-options-object
 - Added i flag to regex to ignore character case
 - Updated logic for makeComment function to accommodate the changes for formatComment

github-actions/issue-labels/add-labels-template.md
```diff
@@ -0,0 +1,18 @@
+<!-- Template for a comment that requires the user to add any missing labels. --->
+
+Hi @${issueCreator}.
+
+Please don't forget to add the proper labels to this issue. Currently, the labels for the following are missing:
+${labels}
+
+To add a label, take a look at Github's documentation [here](https://docs.github.com/en/issues/using-labels-and-milestones-to-track-work/managing-labels#applying-a-label).
+
```
Aveline-art, line 9: I just realized that there is a use case for not using a label (a PM or designer might not know the size of a development issue, for example, and vice versa). In such a case, we would not want to encourage adding a label and hoping for the best, but rather allow someone from the team with the required experience to add the labels instead. For now, let's comment out the function that posts this template (and leave the other template as is) and make a new issue for product to brainstorm the desired behavior and message

Aveline-art (dismissed):

github-actions/issue-labels/add-labels-template.md
Aveline-art: @macho-catt Just comment out the code and this will be ready to merge :) Great work!

macho-catt: @Aveline-art so just comment out lines 5-8 of add-labels-template.md (and accompanying code that maps the labels into this template) until we figure out what PMs want?

Aveline-art: Mmm, do it from the js side, so add-labels-template is never posted in the first place. I don't think the md file needs to be commented out at all?

macho-catt: Ok, so if no labels were added in the first place, don't have the bot comment anything and just add the three missing labels. Is that correct?

Aveline-art: Yep. The missing labels should suffice until the message can be tailored for our optimal workflow.

macho-catt:
### Additional changes

  - Commented out the code where the action will post when the required labels are not added. The action will end (instead of posting something)
  - Added a comment in the code that details why we did this

Aveline-art (commented):
LGTM!

Aveline-art:
The GHA works beautifully! Thanks for making this!

PR #2217, opened 2021-09-03
adding inactive label to in progress issues with no comment or PR by …

### LINKED ISSUE #1977: GitHub Actions: Automate In Progress Issue Updates: Flags Inactive
### Overview
As a developer of the team, it is useful to have automation on the project board for tasks that are done on a regular basis by the leadership team. For this GHA, we will add a '2 weeks inactive' label when an issue has no activity by the assignee for at least 2 week.

### Psudo-code
- [x] Trigger: At the same time as [this GHA](https://github.com/hackforla/website/blob/gh-pages/.github/workflows/add-update-label-weekly.yml)
  - [x] If the issue is in the 'In Progress' column (please use secrets.IN_PROGRESS_COLUMN_ID, which is a value retrieved through the projects and columns api).
  - [x] And if the last comment made by the assignee was at least or over 14 days ago
  - [x] And there is no pull request linked within the last 3 days:
    - [x] Add a the '2 weeks inactive` label

Note: Since this GHA does a related function to [this GHA](https://github.com/hackforla/website/blob/gh-pages/.github/workflows/add-update-label-weekly.yml), this GHA can always live as a separate job on that GHA file, or part of the js code. In any case, since it shares a trigger, it should not live as a separate GHA file. 

### Checks

- [x] If the last comment by the assignee was over 14 days ago
  - [x] and the last comment made was within the last 14 days: flags as inactive
  - [x] and the last action in the [timeline ](https://docs.github.com/v3/issues/timeline/)within the last 14 days is not a comment: flag as inactive
  - [x] and a pull request is linked 2 days ago: **do not** flag as inactive
- [x] If the last comment by the assignee was under 14 days ago: **do not** flag as inactive

### Resources/Instructions

#### Additional resources:

SAUMILDHANKAR:
…assignee in last 14 days

Fixes #1977 

  - Added new functions isInactiveTimelineOutdated and isInactiveMomentRecent to check if the assignee made a comment in last 14 days and PR in the last 3 days.
  - Added inactiveCutoff time to check for assignee activity in last 14 days
  - Reused other functions to update and remove inactive label as required.

<details>

N/A

</details>

<details>
  
N/A

</details>


github-actions/add-update-label-weekly/add-label.js
```diff
@@ -45,6 +50,16 @@ async function main({ g, c }, columnId) {
       await removeLabels(issueNum, toUpdateLabel);
       await addLabels(issueNum, statusUpdatedLabel);
     }
+	
+	// Adds inactive label if the issue's timeline indicates the issue is outdated.
+    if (await isInactiveTimelineOutdated(timeline, issueNum, assignees)) {
+      console.log(`Going to add inactive label for issue #${issueNum}`);
+      await addLabels(issueNum, inactiveLabel);
+    } else {
+      console.log(`No updates needed for issue #${issueNum}`);
```
Aveline-art: I'm not sure this log is needed since the above already posts this log. Consider whether it is preferable to have both, or if one can be safely removed.

github-actions/add-update-label-weekly/add-label.js
```diff
@@ -133,6 +148,27 @@ async function isTimelineOutdated(timeline, issueNum, assignees) {
...
 }
 
+/**
+ * Assesses whether the timeline is outdated.
+ * @param {Array} timeline a list of events in the timeline of an issue, retrieved from the issues API
+ * @param {Number} issueNum the issue's number
+ * @param {String} assignees a list of the issue's assignee's username
+ * @returns true if timeline indicates the issue is outdated, false if not
+ * Note: Outdated means that the assignee did not make a linked PR or comment within the cutoffTime (see global variables).
+ */
```
Aveline-art: The comments do not really describe what this function is about. It sounds like it does the same thing as the one above? In that case, can we combine the functions somehow?

github-actions/add-update-label-weekly/add-label.js
```diff
@@ -133,6 +148,27 @@ async function isTimelineOutdated(timeline, issueNum, assignees) {
...
+async function isInactiveTimelineOutdated(timeline, issueNum, assignees) {
+  for await (let moment of timeline) {
+    if (isInactiveMomentRecent(moment.created_at)) {
+      if (moment.event == 'cross-referenced' && isLinkedIssue(moment, issueNum)) {
+        return false
```
Aveline-art: Is this the check for linked pull requests? If so, the cut off for pull requests is off. The issue calls for a 14 day no-comment cut off AND a 3 day no linked PR cutoff.

github-actions/add-update-label-weekly/add-label.js
```diff
@@ -206,6 +242,16 @@ function isMomentRecent(dateString) {
...
+function isInactiveMomentRecent(dateString) {
+  const dateStringObj = new Date(dateString);
+
+  if (dateStringObj >= inactiveCutoffTime) {
+    return true
+  } else {
+    return false
+  }
+}
+
```
Aveline-art: This function is really similar to the one above. Can the two functions be combined into something a bit more general?

Aveline-art (changes requested):
Hi @SAUMILDHANKAR . I have finished looking over the code. There are a few points that needs work. Please let me know your thoughts!

github-actions/add-update-label-weekly/add-label.js
SAUMILDHANKAR: @Aveline-art Thanks a lot for your comments and review. I will be working on the changes tomorrow after team meeting.

github-actions/add-update-label-weekly/add-label.js
SAUMILDHANKAR: I agree this log is not very helpful and I have removed it.

github-actions/add-update-label-weekly/add-label.js
SAUMILDHANKAR: I have added a new comment which describes the role of this function. It checks if the comment was made more than 14 days ago.

github-actions/add-update-label-weekly/add-label.js
SAUMILDHANKAR: I have modified the code so that it checks for 14 day no-comment and 3 day no linked PR. Thank you.

github-actions/add-update-label-weekly/add-label.js
SAUMILDHANKAR: I have made some changes to these two functions. Please let me know what you think. Thanks.

github-actions/add-update-label-weekly/add-label.js
```diff
@@ -40,13 +45,14 @@ async function main({ g, c }, columnId) {
       await removeLabels(issueNum, statusUpdatedLabel, toUpdateLabel);
       await addLabels(issueNum, toUpdateLabel);
       await postComment(issueNum, assignees);
+    } else if (await isInactiveTimelineOutdated(timeline, issueNum, assignees) && isTimelineOutdated(timeline, issueNum, assignees)) {
+      await addLabels(issueNum, inactiveLabel);
     } else {
       console.log(`No updates needed for issue #${issueNum}`);
       await removeLabels(issueNum, toUpdateLabel);
       await addLabels(issueNum, statusUpdatedLabel);
+	  await removeLabels(issueNum, inactiveLabel);
```
Aveline-art: Not a big deal, per se, but GitHub does have an API call limit, so calling addLabels and removeLabels multiple times somewhat bothers me. Do you think there can be a fix for that (maybe an if else statement) such that api functions are called less? We're not close to the limit at all, so you can feel free to disregard this if this sounds like it might be bigger than the medium issue you were expecting.

github-actions/add-update-label-weekly/add-label.js
```diff
@@ -197,13 +217,21 @@ async function postComment(issueNum, assignees) {
...
+		const dateStringObj = new Date(dateString);
+		if (dateStringObj >= inactiveCutoffTime) {
+			return true
+		}
+		else {
+			return false
+		}
+	}
+	isMomentRecent.isInactiveMomentRecent = isInactiveMomentRecent;
+	return false
```
Aveline-art: Is there a reason why this needs to be a nested function? Both the nested function and the outer function are off by only one variable. Can that variable be refactored as a parameter, and condense this function further?

For example:

```python
func(param1, param2):
  if param1 >= param2:
    return True
  else:
    return False
```

Aveline-art (dismissed):
Great work on getting fixes on this. I have looked at it and it's getting pretty close! I am mostly ask for aesthetic changes because the code needs to be managed in the long run.

github-actions/add-update-label-weekly/add-label.js
Aveline-art: Actually, thinking further, we can probably do something fancy, such as:

```python
func(param1, param2):
  return param1 >= param2
```

Not sure if that is clear, however, so it will need a bit of comments.

SAUMILDHANKAR: Thank you so much for your time on this and valuable review. Sure, I will be starting on the changes.

github-actions/add-update-label-weekly/add-label.js
macho-catt: If you find yourself calling the add / remove github api repeatedly, it might be simpler to use the setLabels function API call. I used it for one of my API calls. You can check /github-actions/issue-labels/check-label.js lines 90-96 for reference. You can also find it in the original doc of [octokit](https://octokit.github.io/rest.js/v18).
You might even be able to use this call and refactor the existing removeLabels + addLabels API calls if we really want to save for github's API call limits.

github-actions/add-update-label-weekly/add-label.js
```diff
@@ -133,6 +139,20 @@ async function isTimelineOutdated(timeline, issueNum, assignees) {
   return true
 }
 
+/**
+ * Note: Here outdated means that the assignee did not make a comment within the inactiveCutoffTime (see global variables).
+ */
+async function isInactiveTimelineOutdated(timeline, issueNum, assignees) {
+  for await (let moment of timeline) {
+    if (isMomentRecent.isInactiveMomentRecent(moment.created_at)) {
+      if (moment.event == 'commented' && isCommentByAssignees(moment, assignees)) {
```
macho-catt: Very small thing, but when using 'equality' operators for javascript, it is often good practice to use strict equality (i.e. ===) to ensure no type conversions between the two things you're comparing. I don't think it matters for this case, but generally it is good practice to follow. Check out this [stackoverflow link](https://stackoverflow.com/questions/359494/which-equals-operator-vs-should-be-used-in-javascript-comparisons) (and other similar articles) for topics on the matter.

github-actions/add-update-label-weekly/add-label.js
Aveline-art: @macho-catt I vaguely recall making this choice because retrieving the labels will require an api call anyways, so keeping tracking of the pre-existing labels was a hassle. That said, there might be another api call here that already retrieves the labels.

github-actions/add-update-label-weekly/add-label.js
SAUMILDHANKAR: Thanks for sharing. Sure, I will keep that in mind.

github-actions/add-update-label-weekly/add-label.js
SAUMILDHANKAR: Removed the nested function.

github-actions/add-update-label-weekly/add-label.js
SAUMILDHANKAR: Made changes to the code logic.

github-actions/add-update-label-weekly/add-label.js
```diff
@@ -18,36 +24,43 @@ cutoffTime.setDate(cutoffTime.getDate() - updatedByDays)
...
+			await addLabels(issueNum, responseObject.labels);
+		} else if (responseObject.result === true && responseObject.labels === inactiveLabel) {
+			console.log(`Going to ask for an update now for issue #${issueNum}`);
+			await removeLabels(issueNum, toUpdateLabel, statusUpdatedLabel);
+			await addLabels(issueNum, responseObject.labels);
+			await postComment(issueNum, assignees);
+		} else {
+			console.log(`No updates needed for issue #${issueNum}`);
+			await removeLabels(issueNum, toUpdateLabel, inactiveLabel);
+			await addLabels(issueNum, statusUpdatedLabel);
```
macho-catt: Change statusUpdatedLabel to responseObject.labels

github-actions/add-update-label-weekly/add-label.js
```diff
@@ -97,41 +110,74 @@ async function* getTimeline(issueNum) {
...
+			}
+		}
+		else if (isMomentRecent(moment.created_at, sevenDayCutoffTime)) { // all the events of an issue between three and seven days will return true
+			if (moment.event == 'cross-referenced' && isLinkedIssue(moment, issueNum)) { // checks if cross referenced between 3 and 7 days
+				console.log('between 3 and 7 cross referenced');
+				return {result: false, labels: statusUpdatedLabel}
+			}
+			else if (moment.event == 'commented' && isCommentByAssignees(moment, assignees)) { // checks if commented between 3 and 7 days
+				console.log('between 3 and 7 commented');
+				return {result: false, labels: toUpdateLabel}
```
macho-catt: Change toUpdateLabel to statusUpdatedLabel

SAUMILDHANKAR: Changed to statusUpdatedLabel

github-actions/add-update-label-weekly/add-label.js
SAUMILDHANKAR: Changed to responseObject.labels

macho-catt:
Hi @SAUMILDHANKAR, I was testing this today and there are still some issues. You can see the logs from my test here: 
https://github.com/macho-catt/website/runs/4543633120?check_suite_focus=true

Specifically, for my own issue number 189 which is currently 2 weeks inactive, it is returning an error: Error: Unhandled error: TypeError: Cannot read property 'result' of undefined

The reason for this is because the last event:
```
  event: 'subscribed',
  commit_id: null,
  commit_url: null,
  created_at: '2021-12-10T07:17:23Z',
  performed_via_github_app: null
}
```
Was created on Dec 10, which will fall under the if statement on line 159 (7-day cutoff time). Since there are no if statements to catch this, the action fails.

I think for the 7-day cutoff time, 14-day cutoff time and the last else statement for anything past 14 days, you need these conditionals as well:
```
			else if (index === timeline.length-1 && (Date.parse(timeline[0].created_at) < fourteenDayCutoffTime.valueOf())) { // returns true if issue was created before 14 days after comparing the two dates in millisecond format  
				return {result: true, labels: inactiveLabel}
			}
			else if (index === timeline.length-1 && (Date.parse(timeline[0].created_at) < threeDayCutoffTime.valueOf())) { // returns true if issue was created before 3 days
				return {result: true, labels: toUpdateLabel}
			}
```

Let me know what you think. Thanks

macho-catt (dismissed):
Please see above comment. Thanks

SAUMILDHANKAR:
@macho-catt: Thanks Mathew for testing this. I was not able to generate this scenario earlier since I was moving the issues to the right column before running the test (and therefore that event would become the latest one). I have added the conditions in all the loops that would make sure it doesn't fail. We can check if it works during the office hours today. 
Also, one workaround for not moving the issue I could think of was to change the column name as well as the secret. 
Thank you.

macho-catt (approved):
LGTM

PR #2223, opened 2021-09-04
Move copy button javascript

### LINKED ISSUE #2101: Refactor Javascript: Move copy button js code to js elements folder
### Overview
As a member of the website development team, Javascript code is much more maintainable and reusable if they exist in a separate folder rather than embedded with html.

### Action Items

- [x] Locate code for copy button [here](https://github.com/hackforla/website/blob/gh-pages/_guide-pages/github-issues.html)
- [x] Move the code to the elements directory in [assets/js](https://github.com/hackforla/website/tree/gh-pages/assets/js)
- [x] Call the file with the copy button with a script tag where the code was originally
- [x] Currently the copy button indicates it was clicked, but never reverts. Fix this behavior such that it returns to its default state. Also, after copying the code to clipboard, the highlighting should disappear.
- [x] Ensure the website appears the same after the changes (except for the copy button).
- [x] **Want to go above and beyond?** You may also try to do something fancy to show transition from default -> clicked state. For example, you might have the text change from "code" to "clicked" and revert back to "code" after a second or so. Feel free to be as creative as you want!

### Resources/Instructions
[assets/js](https://github.com/hackforla/website/tree/gh-pages/assets/js)
[github-issues.html](https://github.com/hackforla/website/blob/gh-pages/_guide-pages/github-issues.html) (contains copy button code)

#### Architecture Notes

The idea behind the refactor is to get all of our website's JS code into an organized, consolidated state. Currently, we want our JS files to exist in our [assets/js ](https://github.com/hackforla/website/tree/gh-pages/assets/js)directory with these specifications:

- In the main js directory, the js files should ideally have names corresponding to a related .html page (for example, updates.js should connect to updates.html).
- In the elements sub-directory, the js fjiles should correspond to code needed for our [design system elements ](https://www.figma.com/file/0RRPy1Ph7HafI3qOITg0Mr/Hack-for-LA-Website?node-id=3464%3A3)to function.
- In the utility sub-directory, the js files represent reusable code that can be used in multiple files. If any js code is reused, they belong in utility, and called in a separate file in the main directory.

That said, this structure is not set in stone. If any part of it feels strange, or you have questions, feel free to bring it up with the team so we can evolve this format!

blakes24:
Fixes #2101

  - Moved the javascript for the copy button to a separate folder
  - Adjusted the CSS so the text would fit inside the button
  - Made button display "copied" then revert to the original state after a short delay

<details>
<summary>Copy button before</summary>

Screen Shot 2021-09-03 at 4 36 20 PM

</details>

<details>
<summary>Copy button after</summary>
  
Screen Shot 2021-09-03 at 4 36 55 PM

</details>

<details>
<summary>Copy button clicked before</summary>

before

</details>

<details>
<summary>Copy button clicked after</summary>
  
copy-btn

</details>


adam-abundis (approved):
Excellent work @blakes24 ! I was able to copy the code from the GitHub issues guide pages with the changes you made. That was a smart trick to wrap the button text in a `<span>` so you can toggle the button text. Keep up the great work! 🥳 🥳 🥳

assets/js/elements/copyButton.js
```diff
@@ -0,0 +1,25 @@
...
+    }, 1000)
+  } catch (err) {
+    const div = document.createElement('div');
+    div.classList.add('copy-err');
+    div.innerText = `Failed to copy`;
+    issueTemplate.append(div);
+    setTimeout(()=> {
+      div.remove()
+    }, 1500)
+  }
```
Aveline-art, line 24: Not a review per se, but I have been curious about the case for the catch block. Under what circumstances do you expect an error to happen?

blakes24: It probably won't ever be an issue since browsers grant permission to the clipboard by default, but navigator.clipboard.writeText() returns a promise and it's possible to get permission denied error when trying to write to the clipboard so I figured it's better to be safe.

adrian-zaragoza (approved):

PR #2252, opened 2021-09-12
Sort closed issues into correct column

### LINKED ISSUE #2206: GHA: Sort closed issues to correct column
### Overview

As a developer, we have to ensure that our kanban board is organized for all teams so that productivity is high. For this issue, we are creating a sorter that will move closed issues and PRs to the correct column by their labels when they are closed.

### Psudo-code
- [x] When an issue/PR is closed:
  - [x] If the issue/PR contains a label from the hard label list:
    - [x] Place the issue/PR in the column marked 'Done'.
  - [x] Else if the issue/PR contains a label from the soft label list:
    - [x] If the issue also contains a label from the soft label counter list:
      - [x] Place the issue/PR in the column marked 'UAT'
    - [x] Else:
      - [x] Place the issue/PR in the column marked 'Done'.
  - [x] Else:
    - [x] Place the issue/PR in the column marked 'UAT'.
- [x] Hard label list: any labels involving refactoring
- [x] Soft label list: 'role: back end' label, 'Feature: Analytics'
- [x] Soft label counter-list: 'role: front end'

### Checks

- [x] An issue containing only 'role: back end' and 'role: front end' labels closes -> UAT.
- [x] An issue containing only 'role: front end' closes -> UAT.
- [x] An issue containing only 'Feature: Refactor CSS', 'Feature: Refactor JS / Liquid', and 'P-Feature: Home page' -> Done.
- [x] An issue containing no label -> UAT
- [x] Make sure the labels are placed in the code in such a way that we can easily add, edit, or delete labels
- [ ] ~~The columns id should ideally be a secret, so that they are not exposed. When the id is found, please consult an admin to add it to [our secrets](https://docs.github.com/en/actions/reference/encrypted-secrets).~~ Columns id can be retrieved as a link, from the columns settings menu, [for example ](https://github.com/hackforla/website/projects/7#column-7198257). Please retrieve the id from there.
- [x] In the PR, request the reviewer to, after merging, test that the board automation does not intercept or supercede this GHA. If so, the team leads will need to remove the default automation to allow this one to work.

Note: The purpose of this issue is to sort closed issues into 2 categories. The "there are visual changes to the website"(C) and "no change"(NC) categories. Anything that is C needs to be in UAT, whereas anything that is NC belongs in Done. If any labels fit the NC category, feel free to tack it onto this issue, but be sure to justify it in your PR.

### Resources/Instructions

#### Additional resources:
[actions/github-script](https://github.com/actions/github-script)
[Encrypted secrets](https://docs.github.com/en/actions/reference/encrypted-secrets)

blakes24:
Fixes #2206

### What changes did you make and why did you make them?

  - Added a GitHub action to sort issues by their labels and move them to the correct column when closed
  - Moves issues with visual changes to UAT and those with no visual changes to Done

### To Review
 - The GitHub action will only work after it is merged
 - After merging, test that the board automation does not intercept or supersede this GHA
 - When testing in my own repo the board automation moved all issues into UAT first, then after about 15 seconds the GHA would move the issue into the correct column


Aveline-art:
This is converted to draft because PM requirements is currently in flux. Until [this comment ](https://github.com/hackforla/website/issues/2206#issuecomment-917758345)has been resolved, this item cannot be reviewed meaningfully or merged.

macho-catt (changes requested):
This is awesome! I tested it in my environment, and it works. Good job. However, I have a couple of comments/change requests regarding the architecture so that it will be easier to update in the future (if needed). Additionally, we are also in the process of refactoring all the github actions, so I want to make that process as smooth as possible.

github-actions/sort-closed-issues.js
```diff
@@ -0,0 +1,57 @@
...
+  // all other issues go to UAT column
+  return "UAT";
+}
+
+/**
+ * Get all labels from the issue
+ * @param {Object} context - context object
+ * @return {Array} - returns an array of all the labels
+ */
+function obtainLabels(context) {
```
macho-catt: It looks like this function is pretty useful and being reused in other github actions. Is it possible to extract this into the github-actions/utils folder? Additionally, can you note the other scripts that use this function so we can make issues out of them? You don't need to change the other scripts that use them.

github-actions/sort-closed-issues.js
```diff
@@ -0,0 +1,57 @@
...
+
+  const issueLabels = obtainLabels(context);
+
+  const includesHardLabel = item => hardLabels.includes(item);
+  const includesSoftLabel = item => softLabels.includes(item);
+  const includesOverrideLabel = item => overrideSoftLabels.includes(item);
+
+  /** if issue includes hard labels there should be no visual changes - move to the Done column */
+  if (issueLabels.some(includesHardLabel)) {
+    return "Done";
```
macho-catt: can "Done" and "UAT" both be put in some kind of const variables at the top? That way, if we ever need to change these (or add more stuff), we can update the variables instead. It'll make it maintainable in the long run.

Aveline-art: Column name strings are viable to change over time. Does the alex-page/github-project-automation-plus@v0.8.1 accept column id rather than column string? Regardless, consider moving the column names into a variable on the top of the file, so that in the scenario that names do change, we only need to perform edits on a single variable.

github-actions/sort-closed-issues.js
```diff
@@ -0,0 +1,57 @@
...
+  ];
+
+  const overrideSoftLabels = ["role: front end"]
+
+  const issueLabels = obtainLabels(context);
+
+  const includesHardLabel = item => hardLabels.includes(item);
+  const includesSoftLabel = item => softLabels.includes(item);
+  const includesOverrideLabel = item => overrideSoftLabels.includes(item);
+
```
Aveline-art: This is simple, but might be hard to read for those weaker with OoP concepts. Consider whether this brevity is worth the a lowered readability.

github-actions/sort-closed-issues.js
```diff
@@ -0,0 +1,57 @@
+/**
```
macho-catt: Can you create a new subdirectory called "move-closed-issues" under 'github-actions' and store sort-closed-issues.js under /github-actions/move-closed-issues? This will make refactoring this action in the future easier.

github-actions/sort-closed-issues.js
```diff
@@ -0,0 +1,57 @@
...
+  const includesSoftLabel = item => softLabels.includes(item);
+  const includesOverrideLabel = item => overrideSoftLabels.includes(item);
+
+  /** if issue includes hard labels there should be no visual changes - move to the Done column */
+  if (issueLabels.some(includesHardLabel)) {
+    return "Done";
+  }
+
+  /** if issue does not include a hard label, but does contain an override label - move to UAT */
+  if (issueLabels.some(includesOverrideLabel)) {
```
macho-catt: Is there a reason why you're specifically using multiples ifs here, instead of an if/else conditional? I know the results are mutually exclusive so it really doesn't matter which one you use, I just want to know your reasoning.

Aveline-art (dismissed):
Hi @blakes24 . Just a couple of comments for this one! Please let me know your thoughts!

github-actions/sort-closed-issues.js
blakes24: It's just a style preference. Since all the ifs have returns using else seems redundant, and it just feels more readable to me this way than a chain of else ifs, but I can change it to an if-else if you prefer.

github-actions/sort-closed-issues.js
blakes24: Will do

github-actions/sort-closed-issues.js
blakes24: Done. It looks like the only other file using it is `issue-labels/check-labels.js`

github-actions/sort-closed-issues.js
blakes24: The action doesn't accept column ids, but I have assigned the column names to variables.

github-actions/sort-closed-issues.js
blakes24: Would it be more readable if I made them standard functions instead of arrow functions?

Aveline-art: I am not exactly sure how it can be more readable, but stray observations:

1. Each function basically accomplishes the same thing.
2. Is this the best way to create these helper functions? What if requirements change again?
3. The main reason why this is hard to read is because it is not immediately clear the purpose of these lines, and why there needs to be a 121 correspondence with each array rather than one function that takes in that array.
4. A function that returns a function might be briefer?
5. Comments?

blakes24: Would something like this be clearer?
```
/** Returns a function that checks if a label is in the given array */
function checkLabels(labelArray) {
    return (label) => labelArray.includes(label);
  }
```
then it would be used like this:
```
if (issueLabels.some(checkLabels(hardLabels))) {
    return doneColumn;
  }
```

Aveline-art: I'm not too sure if this is clearer per se... Let's abandon this line for the time being, and just add clarifying comments.

.github/workflows/move-closed-issues.yaml
```diff
@@ -0,0 +1,24 @@
+name: Move Closed Issues
+on:
+  issues:
+    types:
+      - closed
+jobs:
```
Aveline-art, line 6: Hi @blakes24 . This is not a fault of your own, but we want to transition our GHAs into something more manageable. Specifically, we want to move all issue triggers into [one file](https://github.com/hackforla/website/blob/gh-pages/.github/workflows/issue-trigger.yml), with if else clauses in the file differentiating different types of triggers. @macho-catt, please clarify and confirm if @blakes24 should transition this to our issue mega file.

macho-catt: I am okay with approving this and creating a new issue to refactor into that file. It might add more work, but I probably will take the issue and put another job in that file first (add-issue-labels) so other developers can just look at it and know how the structure with if conditionals will work.

macho-catt (approved):
Worked in my environment. Good job!

Issue #2385, opened 2021-10-20
Google Apps Script: Add unique IDs as object keys for each wins entry


macho-catt:
### Dependency 
- [x] #3722
- [x] #3641
- [x] #4134
- [x] #2147

### Overview
To expedite the process of locating a specific wins entry, each wins entry (which is also a javascript object) should have a unique ID as a key.

### Action Items
- [x] Follow sections 1-3 in the Wins Apps Script Development Process[^3].  Note that in this issue you will be modifying both Apps Scripts `wins-form-responses` and `gh-requests` so setup will be required in both `/google-apps-script/wins-form-responses` and `/google-apps-script/gh-requests` 
- [x] As described in section 3a and 3b, you will provide a dev lead with the google account you wish to use for testing, and request the dev lead to:
   - do not modify folder level access
   -  share the Wins Form Admin Guide[^1] as VIEWER 
   -  share the Wins-form (Response)[^2] sheet as VIEWER
   -  share the gh-requests doc as VIEWER
- [x] Continue following instructions in 3c and d.  As described in 3d, you will have to make modifications to certain documents in the test folder. 
   - [x] It will not be necessary to restore any triggers, because you can execute `main()` in debug mode. 
   - [x] Since you will be running `main()` for testing, you must follow instructions in [Testing WINS in a test repository](https://github.com/hackforla/website/issues/5746#issuecomment-1846549069)
- [x] Continue following instructions through section 6 "Editing".  As described in this section, editing can be done locally in an IDE or in the Google Drive cloud editor, and the `clasp` utility can be used to transfer code back and forth between the local worksation and the Google Drive as needed. 
    - [x] Modify wins so that on the creation of the javascript object for each entry, a unique ID is generated as well
    - [x] This ID can be the row number of the wins entry
- [x] Ensure that even with the addition of the unique IDs as keys, the current wins page should functionally look and work the same way in your own repo. 
- [ ] Demo your changes to the website leads so they can approve and add it to the production script

### Resources/Instructions
[Google Apps Script](https://developers.google.com/apps-script/overview)
[_wins-data file](https://github.com/hackforla/website/blob/gh-pages/_data/external/_wins-data.json)
[wins page JS](https://github.com/hackforla/website/blob/gh-pages/assets/js/wins.js)
[^1]:[Wins Form Admin Guide](https://docs.google.com/document/d/1j1JTX5XzotgVCVZ91ImvenedPN5-IVjdSKDukq1B9iE/edit?usp=drive_link)
[...]

PR #2623, opened 2021-12-23

elizabethhonest:


JessicaLucindaCheng:
After discussing this pr with @SAUMILDHANKAR, we have decided to close this pr because there are commits made from testing issue #2385 along with actual bot commits and since there are so many commits, it's hard to tell which are from testing and which are actually from the bot itself. We can always manually trigger a new pr so we can incorporate the new wins.

PR #3741, opened 2022-11-22
Update docker-compose.yml to clean Jekyll data

### LINKED ISSUE #3587: Update docker-compose to remove the previously generated site and Jekyll metadata before building
### Overview

Jekyll does not always rebuild every file in the _site directory every build. This can cause old versions of files to be served.

### Action Items

- [x] Update the `docker-compose.yml` file to run the `jekyll clean` command before the existing `jekyll serve` command.
- [x] Test updating a file in `_projects` to ensure the update reflects immediately when running `docker-compose up`.

### Resources/Instructions

- https://github.com/hackforla/website/blob/gh-pages/docker-compose.yml
- https://jekyllrb.com/docs/usage/

Adastros:
Fixes #3587 

Updated `docker-compose.yml` file to add a Jekyll command (`jekyll clean`) to the command key. The `jekyll clean` command removes all of the generated site and Jekyll metadata before docker builds the site. Without the `jekyll clean` command, Jekyll may not rebuild every file and serve old files instead. 
  
In addition to the `jekyll clean` command, `sh -c` was added in order to run multiple Jekyll commands. `sh` indicates which language interpreter to use and the `-c` argument indicates to read the jekyll commands in the following string. The `&&` part of the string separates the jekyll commands. `sh` was chosen over `bash` since `sh` is more widely supported. 

`>` was also added to indicate that the value for the command key is a multi-line string and that newlines should be treated as spaces when read. 

### Screenshots of Proposed Changes Of The Website 
No visual changes to the website.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -3,7 +3,9 @@ services:
   hfla_site:
     image: hackforlaops/ghpages:latest
     container_name: hfla_site
-    command: jekyll serve --force_polling --livereload --config _config.yml,_config.docker.yml -I
+    command: >
+      sh -c "jekyll clean &&
+             jekyll serve --force_polling --livereload --config _config.yml,_config.docker.yml -I"
     environment:
       - JEKYLL_ENV=docker
     ports:


roslynwythe (approved):
Tested on Windows machine, running docker-compose up from cmd.exe and also from git bash shell, worked fine.   The cleaning does lengthen the time to bring up the site, from 26 to 93 sec.  Thank you for explaining the method for specifying multiple jekyll commands, as I'd never seen that before.  Great work.

blulady:
It worked for me as well, also a windows user. We should wait for a review from someone with a Mac. I am also thinking about changes I can make to test the code to make sure that it is functioning as designed. The explanation is great @Adastros!! Do you mind sharing some resources?

jdingeman (approved):
Also tested on a Windows machine. Though the build time is longer, I believe it's better in the long run so that we don't run into any issues with Docker serving old images of the site. Like Sarah said, I'd like to see a Mac OS user chime in on this to make sure it runs fine on their machine.

Adastros:
Hi @blulady, I don't mind sharing! Apologies in advance if there's a bit of information overload. Below are the resources I consulted while working on the issue. 

I also want to add that I tested this on Windows. I did my best to choose the command with the widest support since I wasn't able to test this on Linux or macOS. 

### Jekyll
- https://jekyllrb.com/docs/usage/ 
- https://jekyllrb.com/docs/structure/

### docker-compose multi-line (`>`)
- https://dev.to/serhatteker/how-to-run-multiple-long-commands-in-docker-compose-1f13
- https://stackoverflow.com/questions/3790454/how-do-i-break-a-string-in-yaml-over-multiple-lines
- https://yaml-multiline.info/

### `sh`, `-c` , `&&`, `bash`
- https://stackoverflow.com/questions/5725296/difference-between-sh-and-bash 
- https://linuxhint.com/compare-sh-shell-vs-bash/
- https://unix.stackexchange.com/questions/555099/what-is-the-most-portable-shell-and-relevant-best-practices-to-follow
- https://askubuntu.com/questions/831847/what-is-the-sh-c-command#:~:text=sh%20calls%20the%20program%20sh,the%20command%20instead%20of%20bash.
- https://pubs.opengroup.org/onlinepubs/9699919799/idx/shell.html

roslynwythe:
@Adastros I have a question.   Does your new command run if the user stops and starts the website container from the Docker Desktop application?

MattPereira (approved):
The updated Jekyll command works as intended on Linux :penguin: 

 The build time jumps from 3.3 to 9.9 seconds, but like Justin said, it's a worthwhile tradeoff. 

@Adastros outstanding job explaining all of your changes for this PR! :+1:

Adastros:
Hi @roslynwythe, based on my understanding of the `start` and `stop` commands for Docker the `jekyll clean` command won't run. The `start` and `stop` commands only involve the containers for the website. Everything else that's needed to host the website locally won't be touched by the `start` and `stop` commands. The `start` and `stop` commands are kind of like a soft start and stop (if that makes sense lol). `docker compose up` and `docker compose down` are more like hard start and stops. Those commands start and stop all resources needed to run the website. That being said, in order to run the website locally we would need to use `docker compose up`. Doing so would run the `jekyll clean` command added with this pull request.

PR #3942, opened 2023-02-06

elizabethhonest:


jdingeman:
I noticed that the section used to update `wins-data.json` in the Apps Script uses `Utilities.Charset.UTF_8` after a dev attempted to fix the issue with the question marks showing up in the Wins page where apostrophes are supposed to be. 
image

I added that parameter to the `_wins-data.json` section because I noticed that the `/assets/js/wins.js` calls on `_wins-data.json` to display the wins on the page. I'm using this PR to test the change.

Issue #4530, opened 2023-04-17
ER: Filters on project page


JessicaLucindaCheng:
### Emergent Requirement - Problem
On the current homepage and Projects page, I can't see all the filter categories unless I do one of the following:
- Scroll all the way through all the projects
- Or collapse the Program Area filter in order to see the Status filter category or be able to see the `View all` in the Languages/Technologies filter category

To see the problem I am seeing, view the https://www.hackforla.org/projects/ page using Developer Tools and inspect it at 1024 x 659 and look at the filters on the left side while scrolling down the page.

<details>
  <summary>Click here to see the problem with not being able to see all the filters while scrolling</summary>
  <img src="https://user-images.githubusercontent.com/31293603/234155930-b1dabfdb-1c47-4164-9df4-ccee4e40fab1.gif" alt="">

</details>

### Issue you discovered this emergent requirement in
- Not applicable

### Date discovered 
2023-Apr-17

### Did you have to do something temporarily
- [x] YES
- [ ] NO

Scroll all the way through the projects or collapse a filter category

### Who was involved
@JessicaLucindaCheng 

### What happens if this is not addressed
It's not user friendly

### Resources
- https://www.hackforla.org/projects/

### Recommended Action Items
- [x] Discuss with Stakeholder Rep (Bonnie) potential solutions
- [x] Make a new issue
- [x] Discuss with team

ExperimentsInHonesty:
We should make a new issue to fix it so that it operates the same way as REI and Amazon (pined to the top)

Also filter should be pinned to content
![image](https://user-images.githubusercontent.com/37763229/234152975-f524a3c0-d9a1-4edc-a84b-f0672c1b9198.png)
![image](https://user-images.githubusercontent.com/37763229/234153027-bd5488af-08fd-406c-b522-2277be8d6ad0.png)

chrismenke45:
Availability: 8am-12pm weekdays
ETA: 4/28 EOD

Adastros:
Availability (PST): 10 - 11:30 PM 03MAY2023
ETA (PST): 11:30 PM 03MAY2023

Adastros:
- Issue for ER can be found in issue #4593

Adastros:
Issue is done drafting and ready for prioritization. Closing ER

Anahisv23:
[...]

PR #4694, opened 2023-05-18
"Projects" and "Home Page" Edits | pins project filter to top of project list, un-sticky's the Filters on scroll, & centers components

### LINKED ISSUE #4593: Projects and Home Pages: Pin Projects Filter to the Top of the Projects List
### Overview

We need to improve the usability of the filters, search and project display so that it is easy for people to navigate the project list.

#### Details
- Make the projects filter stay at the top of the content section of the page and scroll independently to allow the user to scroll through all filter options and projects at the same time.  So while scrolling through the matching projects on the right, the project filters do not change position".  

See how the [REI website](https://www.rei.com/c/sneakers) implements their page when scrolling through the merchandise on the website. The projects filter and list should look and behave similar to it.

The difference between REI and ours is that we have a search bar, that will stay pinned at the top.

### Action Items
- [x] Update the styling for the filter so that the filter is anchored to the top of the parent section element 
- [x] Update the styling for the section element containing the filter and projects list so that the two components scroll independently
  - [x] are pinned to each other when the viewport changes for viewport widths above 767px
  - [x] are centered to the page when the viewport changes for viewport widths above 767px

### Resources/Instructions

- [Hack for LA home page](https://www.hackforla.org/)
- [Hack for LA projects page](https://www.hackforla.org/projects/)
- [_dropdown_filters.scss](https://github.com/hackforla/website/blob/gh-pages/_sass/elements/_dropdown_filters.scss)
- [ER Request for this issue](https://github.com/hackforla/website/issues/4530)

agosmou:
Fixes #4593 

  - Pinned the projects filters to the top of the projects list to mimic the REI website
       - UX/UI best practice to facilitate user navigation along the page
  - Removed the "sticky" styling for the filters side pane
       - This caused unnecessary scrolling through all project cards to release the filters pane so the user can see the filters at the bottom of the pane
  - Centered the components to the center to mimic the REI website
     - having the components centered can be considered good UX/UI since the user can quickly navigate because there is less distance between components
- Refactored where needed 
 

<details>

image

</details>

<details>

image
  
image

</details>

### DIFF
--- a/_includes/current-projects-check.html
+++ b/_includes/current-projects-check.html
@@ -1,6 +1,6 @@
 <a class="anchor" id="projects"></a>
 <section class="content-section projects">
-  <div class="page-contain projects-inner">
+  <div class="page-contain projects-inner curr-projects">
     <h2 class="project-header title2">Current Projects</h2>
     <p class="project-pitch">
       Have an idea?
@@ -13,7 +13,7 @@
     <ul class="filter-list" id="filter-list"></ul>
     <div class="filter-tag-container"></div>
   </div>
-  <div class="page-contain projects-inner" style="clear: left">
+  <div class="page-contain projects-inner curr-projects" style="clear: left">
     <ul class="project-list unstyled-list"></ul>
   </div>
 </section>
--- a/_includes/current-projects.html
+++ b/_includes/current-projects.html
@@ -31,7 +31,7 @@
     </div>
   </nav>
   <div class="projects-and-filters">
-    <div class="filter-tag-container"></div>
+    <div class="filter-tag-container top-filter"></div>
     <div class="page-contain projects-inner" >
       <ul class="project-list unstyled-list"></ul>
     </div>
--- a/_includes/current-projects.html
+++ b/_includes/current-projects.html
@@ -18,7 +18,7 @@
           </svg>
         </button>
       </h3>
-      <div class="filter-tag-container"></div>
+      {% comment %} <div class="filter-tag-container"></div> {% endcomment %}
       <ul class="filter-list" id="filter-list"></ul>
       <div class="mobile-filter-buttons">
         <button class="cancel-mobile-filters btn-md btn-dark">
@@ -30,8 +30,11 @@
       </div>
     </div>
   </nav>
-  <div class="page-contain projects-inner" style="clear: left">
-    <ul class="project-list unstyled-list"></ul>
+  <div class="projects-and-filters">
+    <div class="filter-tag-container"></div>
+    <div class="page-contain projects-inner" >
+      <ul class="project-list unstyled-list"></ul>
+    </div>
   </div>
 </section>
 
--- a/_includes/current-projects.html
+++ b/_includes/current-projects.html
@@ -1,7 +1,7 @@
 <a class="anchor" id="projects"></a>
 <section class="content-section projects filter-content-container">
   <nav class="filter-toolbar" aria-label="Filter Navbar">
-    <div class="stickyDiv">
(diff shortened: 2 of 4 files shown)


t-will-gillis:
Availability: 5/18 evening, 5/19
ETA: 5/19

mademarc:
Review ETA: 5/18/2023
Availability: 7:05 PM

mademarc (dismissed):
Hey @agosmou  great adding the filter on the div class to keep the code very well read.

t-will-gillis (changes requested):
Hey @agosmou great job adjusting the projects filter: removing the sticky styling, pinning to the top of the project list, and centering on the page for the Projects and homepages. 

When I view in Docker, I am noticing the following:
- In all views, the 'Applied Filters' notice no longer appears. I am not arguing whether or not the 'Applied Filter' notice is useful or worth keeping, only that removing this notice is beyond the scope of the issue.
- In tablet and desktop views, the red applied filter tags have shifted from the Filters column and now are above the project cards. I believe that keeping the red applied filter tags in the current location is more intuitive for the user.
- In the tablet/ 767 px view, the Filters column width compresses and looks 'smushed' in comparison to the way it looks currently in the same view. I feel that maintaining the width of the Filters column is preferable visually.
comparison_767
- Lastly, I am not entirely clear about how the [/projects-check](https://www.hackforla.org/projects-check/) page is used currently or whether there is even a link to this page from the current website, however the edits made with the PR also change the layout of this page- both the 'Current Projects' header and the project cards themselves become off-center. 
comparison_projects-check
Thanks for your work so far!

agosmou:
Hey @t-will-gillis - Thank you for your thoroughness. See Below:

***

> * In all views, the 'Applied Filters' notice no longer appears. I am not arguing whether or not the 'Applied Filter' notice is useful or worth keeping, only that removing this notice is beyond the scope of the issue.

I feel like this was implied as I had to match REI's page (where their filter's don't have that text) per [Bonnie's comment on the Emergent Request](https://github.com/hackforla/website/issues/4530#issuecomment-1521032707)

> * In tablet and desktop views, the red applied filter tags have shifted from the Filters column and now are above the project cards. I believe that keeping the red applied filter tags in the current location is more intuitive for the user.

I had to match REI/Amazon's page per [Bonnie's comment on the Emergent Request](https://github.com/hackforla/website/issues/4530#issuecomment-1521032707)

> * In the tablet/ 767 px view, the Filters column width compresses and looks 'smushed' in comparison to the way it looks currently in the same view. I feel that maintaining the width of the Filters column is preferable visually.

Agreed! I adjusted this.
image


> * Lastly, I am not entirely clear about how the [/projects-check](https://www.hackforla.org/projects-check/) page is used currently or whether there is even a link to this page from the current website, however the edits made with the PR also change the layout of this page- both the 'Current Projects' header and the project cards themselves become off-center.

Nice catch - Those properties are used in a lot of places throughout the whole website. I went ahead and fixed this in the CSS just in case (not sure this page is used).
image

t-will-gillis (approved):
Hey @agosmou  – Thanks for the clarification/ explanation about the placement of the red applied filter tags. I can see why you did what you did when I look at the REI page. Also, when the filter tags are at the top of the project cards, there is not really a good location for the ‘Applied Filters’ verbiage and since it does not add to clarity, I agree with you that the text can/should be removed.

rei_comp

Thanks also for changing the min width of the Filters column (I am just noticing the responsive design changes from 1 to 2 to 3 columns going from 766 to 767 to 768 on the current site) and for adjusting the /projects-check page. Both look good now.

Great job working through this issue!

PR #4942, opened 2023-07-12
Add a search function to projects page- Draft

### LINKED ISSUE #4136: Add a search function to the projects page
### Overview
The [projects page](https://www.hackforla.org/projects/) needs a search box so that a user can search all of the projects by keywords. This issue is to implement the final design for the search box. 

### Action Items
- [x] Add a search box at the top of the projects to reflect the design in Figma
- [x] Similar to a user using ctrl + f to find a word on a webpage, the user can search any word or words from the descriptions of the projects
  - [x] The difference would be that the search function will narrow down the projects rendered on the page based on user input
- [x] Include functionality for [boolean operators](https://southern.libguides.com/google/boolean) for users to refine their search
  - [x] Only use operators AND, OR and - 
   <html>
<body>

Boolean Operators | Meaning | Example
-- | -- | --
**AND** | Limit results | Vegetarianism AND obesity(Search for content that contains both vegetarianism and obesity.)
**OR** | One term OR another | Scholarly OR academic(Search for content that contains scholarly OR academic.)
**-** | Exclude a term from the search | Bears -Chicago(Limits results to only those with bears and not the term Chicago.)

</body>
</html>

- [x] Make the appropriate changes to the CSS, HTML, and JS as you need to
- [x] The goal is to make this component reusable on multiple pages so it can be used for current and future pages
- [x] Once you have submitted the PR, please write up what files go into the search bar component and any other details that a person adding the same component to another page (e.g., toolkit page) will need to know. You can leave that write up here in this issue as a comment.

### Resources/Instructions
- See the [final figma design](https://www.figma.com/file/0RRPy1Ph7HafI3qOITg0Mr/Hack-for-LA-Website?node-id=24096-281035&t=jNu8aChcchEtRwmG-0)
   - see [toolkit mobile figma latest](https://www.figma.com/file/0RRPy1Ph7HafI3qOITg0Mr/Hack-for-LA-Website?type=design&node-id=24535-287271&mode=design&t=f7zbZLTg6r3V9B07-0) for search inspo 
- In the [_includes](https://github.com/hackforla/website/blob/gh-pages/_includes/) page you should find the [current-projects.html](https://github.com/hackforla/website/blob/gh-pages/_includes/current-projects.html)
- In the [assets/js](https://github.com/hackforla/website/blob/gh-pages/assets/js) folder you should find [project.js](https://github.com/hackforla/website/blob/gh-pages/assets/js/project.js)
- [Standardized Components](https://github.com/hackforla/website/wiki/Standardized-Components) wiki
-

Thinking-Panda:
Fixes #4136

  - Edited html and css files to insert a search bar. Imported the `_search-bar.scss` in `main.scss`
  - Edited `current-projects.js` so the search function displays the desired result.
  

  - Html changes- added the search bar as a form
  - CSS changes- added a new file with style properties for the search bar and imported it in `main.scss`
  - JS changes- added functions for click events in the search bar. Edited the function `updateProjectCardDisplayState(filterParams)` to incorporate results from the search function

<details>

beforeChangeMobile

beforeChangeTablet

beforeChange

</details>

<details>
  
AfterChangeMobile
AfterChangeTablet
AfterChangeLaptop
search+filterResults

</details>


Thinking-Panda:
@kwangric - Let me know if you have any suggestion of code change to combine the filter and search results.

kwangric (commented):
Issue resolved over Slack.

roslynwythe:
Hi @Thinking-Panda if you were able to get the help you needed, please remove the `Help Wanted` and `Status Updated` labels.  Thank you @kwangric for assisting with the code.

Thinking-Panda:
@Merge Team- I have made the necessary code changes and this PR is ready to be reviewed and merged. Thanks!

kwangric:
ETA: EOD 8/5
Availability: Most days after 3pm PT

kwangric (changes requested):
@Thinking-Panda, great job with this issue!
- The search bar and function works as expected. Projects are correctly filtered based on whether the search term is included in the description or not.
- The search works in conjunction with the filters.
- Boolean operators can be used to refine the search.
- Search bar is also displayed and works correctly in mobile view.

I just have two requests for changes to be made:
- You used `NOT` instead of the `-` symbol for the exclude operator. Given that the issue and the link to boolean operators included uses `-`, it might be best to use that instead.
- For the booleans operators, I noticed that the operators work even when not in uppercase. This might cause confusion fs the user is searching for terms that include the word 'and' & 'or', but aren't trying to use the operators. Can they be changed so they only work if they are in uppercase?

Thinking-Panda:
@kwangric , the requested changes have been done. Please review. Thanks.

LRenDO:
Hi @mademarc! Please add your ETA and Availablity when you have a minute.  Thanks!

kwangric (changes requested):
Thanks for the a quick turnaround on the changes Niki!

- Boolean operators are now working only when in uppercase
- The operators use `-` instead of `NOT`

One last small request: at the moment in order to exclude a term, you will need a space between the `-` symbol and the search term (e.g. `HTML - react`). I believe there shouldn't be a space for the operator to work (`HTML -react`). Can you make this change? The other operators work as expected.

Thinking-Panda:
@kwangric , @mademarc - Change requested has been implemented. Thank you for your patience. The PR is now ready to be re-reviewed.

assets/js/current-projects.js
```diff
@@ -246,6 +268,40 @@ function cancelMobileFiltersEventHandler(e) {
...
+            }
+        }  
+    }
+    let queryString = Object.keys(tokenObj).map(key => key + '=' + tokenObj[key]).join('&').replaceAll(" ","+");
+
+    window.history.replaceState(null,'',`?${queryString}`)
+}
+
+function searchEnterKeyHandler(e){
+    if (event.key === "Enter") {
```
StephenTheDev1001: Same here. The [event property](https://developer.mozilla.org/en-US/docs/Web/API/Window/event) is deprecated.
```suggestion
    if (e.key === "Enter") {
```

assets/js/current-projects.js
```diff
@@ -246,6 +268,40 @@ function cancelMobileFiltersEventHandler(e) {
...
+        document.querySelector(".search-glass").click();
```
StephenTheDev1001: Instead of using a querySelector that calls searchEventHandler, I think it'd be more efficient just to call searchEventHandler().
```suggestion
       searchEventHandler(e);
```

assets/js/current-projects.js
```diff
@@ -246,6 +268,40 @@ function cancelMobileFiltersEventHandler(e) {
...
+    }
+}
+
+function searchOnFocusEventHandler(){
+    document.querySelector(".search-x").style.display='block';
+}
+
+function searchCloseEventHandler(){
```
StephenTheDev1001: Same here. The [event property](https://developer.mozilla.org/en-US/docs/Web/API/Window/event) is deprecated.
```suggestion
function searchCloseEventHandler(e){
e.preventDefault();
```

assets/js/current-projects.js
```diff
@@ -246,6 +268,40 @@ function cancelMobileFiltersEventHandler(e) {
     hideFiltersEventHandler(e)
     clearAllEventHandler()
 }
+//search bar event handler
+function searchEventHandler(){
```
StephenTheDev1001: The [event property](https://developer.mozilla.org/en-US/docs/Web/API/Window/event) is deprecated. You can pass the event as a parameter.
```suggestion
function searchEventHandler(e){
e.preventDefault();
```

StephenTheDev1001 (commented):
The search bar looks great, pixel perfect implementation from the designs. The search bar works as intended. I'm not sure how to handle the [event property](https://developer.mozilla.org/en-US/docs/Web/API/Window/event)  here. Would it be better to do it the old deprecated way to be consistent or use the newer way to avoid errors in newer browsers? Also a small suggestion for searchEnterKeyHandler.

It's obvious you put a lot of time and effort into this. Thank you.

@jdingeman @roslynwythe How do you recommend the [event property](https://developer.mozilla.org/en-US/docs/Web/API/Window/event) be handled?

mademarc:
Review ETA: 8/11/2023
Availability: 5:24PM

mademarc (dismissed):
Hey @Thinking-Panda  the html elements were added very well, same with the scss file and the js file with proper formatting as well and @LRenDO the ETA and Availability is done.

kwangric (dismissed):
Thanks for making the requested changes, the search bar is working as intended and everything looks good visually. Great work!

assets/js/current-projects.js
StephenTheDev1001: @Thinking-Panda do you think this would cause problems doing it this way

Thinking-Panda:
@StephenTheDev1001, @kwangric @mademarc  , suggested changes have been made and the function still works good. Please re-review

StephenTheDev1001 (approved):
The search bar works as intended. Good job with the requested changes.

kwangric (approved):
The search function still functions as expected. 

One thing I noticed in previous commits was that the text in the search bar wouldn't clear when you click the magnifying glass to search, while pressing enter would clear the text. The latest change seems to keep the text from clearing when hitting enter, which is good for consistency sake.

I'm not familiar with the Lint SCSS test, so I'm not sure what is causing it to fail the check. I will see if anyone else on the merge team can help resolve it.

kwangric:
It looks like the failed check is an issue with the linter itself. Since everything else looks good, I will go ahead and merge the PR. Great job with the issue @Thinking-Panda!

PR #4950, opened 2023-07-15
feat: automation for moving ice box items to new issue approval column

JpadillaCoding:
Fixes #3732

### What changes did you make?
Made a workflow to move Ice Box items to the New Issue Approval column under the following circumstances:
  - In the dependency section, all enumerated action items are checked off
  - Cannot contain action items that are comments 

### Why did you make the changes (we will use this info to test)?
  - To increase productivity, issues with completed dependencies from the Ice Box are automatically moved to the New Issue Approval column along with adding the "ready for dev lead" label
  - The script checks for the dependency section's action list. Since a Template is for the issue approval process, the script looks for "### Dependency".
  -  A separate check is used to verify all action items are linked to an issue using regex since they're made using markdown.
  - If there are no action items in this section then the workflow is aborted since there is nothing to check for.
  - If all of the above pass the the issue is moved to the new issue approval column & the "ready for dev lead" label is added to streamline the issue approval process. 
  - A local secret token had to be made for local testing. This has been changed to use the hack for la token as seen on other workflow files.

### Screenshots of Proposed Changes Of The Website  (if any, please do not screen shot code changes)

<details>
<summary>Demo Of Workflow in local enviroment</summary>

issue_mover_demo

Three test were made to follow the testing guidelines. 
"Pass Tester" contains only an enumerated action list and should pass to moving to the new issue approval column.
"Fail Tester" contains no list and only a comment. This test should fail. 
"Fail Tester2" contains an action list with enumerated and comment action items. This test should fail.

</details>

### Test Performed 

I am including the process I used to test this github action so that it can be replicated on review.

- [ ]  Copy the project board into a test repo. It is important to do this step since Hack For La uses the classic project board. Making a classic project is no longer an option unless it is copied. This [Link](https://docs.github.com/en/issues/organizing-your-work-with-project-boards/managing-project-boards/copying-a-project-board) shows how to do that.
- [ ] Check for a "ready for dev lead" label. If the copied project does not have one, then one should be made.
- [ ]  A personal access token has to be made for the project. Make sure to save the generated code for the next step. This [Link](https://docs.github.com/en/enterprise-server@3.6/authentication/keeping-your-account-and-data-secure/managing-your-personal-access-tokens) shows how to do that. Make sure the following permissions are checked "Repo", "Workflow", and "Project".
- [ ] Make a repository secret by going to the settings in the test repo. 
   - [ ] Click on "Secrets & Variables". 
   - [ ] Under this tab click "Actions". 
[...]

PR #5053, opened 2023-07-23
Revert "feat: automation for moving ice box items to new issue approval column"

roslynwythe:
Reverts hackforla/website#4950.  Bug reported in which issues from columns other than the "Ice Box" are moved.


t-will-gillis (approved):

JpadillaCoding:
Workflow had the potential to be triggered on every column on the project board. 
- This was causing a bug where any issues being edited could be potentially affected if it has a "Dependency" section. If all action items in this section are checked off then the issues would be moved to the "New issue approval" column with the "ready for dev lead" label.

There was a misunderstanding on how the "filter" keyword works. An alternative needs to be found to target only the "Ice Box" column in the "project board" project.

JpadillaCoding:
It seems there is no native way to move issues based on a specific column. 

I attempted to make my own script by using the Github API to pull object information on the issue being worked on and the target column, but I end up in the same spot of not being able to connect an issue to a specific column. 
The Issue endpoint from the API does not contain relational information related to github projects. 

The only solution I can think of is using [this](https://github.com/marketplace/actions/move-labeled-or-milestoned-issue-to-a-specific-project-column) action from the marketplace that moves an issue to another column when a specific label is added. It would just be a change of scope from the original issue. 

I'd be happy to show the progress I made if needed.

Issue #5639, opened 2023-10-01
Review Inactive Members


aramattamara:
**Inactive Members:**

cnk
vincentdang
angelenelm
DeviantSchemist
Sebster3
averdin2
aadilahmed
Karll-L
abuna1985
drakenguyen4000
bluechocolate2019
Toe12
Tomomi-K1
kennytram
julialuongw
muninnhugin
jennifertieu
agutiernc
kwangric
arfgit
robertnjenga
steven-positive-tran
ArchilChovatiya
DorianDeptuch
HackforLABot
DomenicScozz
SeyiAkinwale
MattPereira
vanyanv
k-rewd
whitneywind
oliviazhai
blaycoder
pdimaano
ericvennemeyer
bphan002
Abrosss
junadkat32
Josiah-O
LukeLowrey2
elizabethhonest
ziniwang
thedvo
Jaretzbalba
jch1013
ronaldpaek
erinzz
kiwookim
matthewmpan
93Belen
aramattamara
aidenshaw07
ldietz08
LOSjr4
sijiapitts
JpadillaCoding
tesiahwang
RyanGehris
ortegaa32
lateral-neck-stretch
bootcamp-brian
Vinny02
allison-truhlar
garrettallen0
DevRishiJain
Thinking-Panda
evanhesketh
JanineSooThow
jnomad21
benpinhassi
hfla-website-checklist
Jmmcclo2023
tkozek

ExperimentsInHonesty:
I added this issue to the project board and added the milestone team workflow during a label audit

t-will-gillis:
Hello @aramattamara This issue was created from your repo because the “Schedule Monthly PREV” GitHub Action ran automatically and wrote the issue to the HfLA Project Board. The only way for this to happen is if you have created a “Repository secret” and used the same name as the HfLA token referenced in the action that is running- in this case `HACKFORLA_BOT_TOKEN`. 
  
Please delete this and any other Repository secrets (the secrets you have authorized in your own repo) that share the same name as an HfLA token. As I mentioned in the issue you were reviewing, next time please give your repository secret/ token any other name- for example `TEST_TOKEN`  - as long as it is something different from the HfLA token.  

![Screenshot 2023-10-01 153409](https://github.com/hackforla/website/assets/40799239/4e8cedfa-e54c-4749-83b7-11f764e37e90)

aramattamara:
Sorry, now I realized how it works. Thank you! Removed

PR #5467, opened 2023-09-09
Fix bug and consolidate "Schedule Thursday" into "Schedule Monthly" 4788

t-will-gillis:
Fixes #4788 

### Notes to PR Reviewers
  - You will need to have a functioning test environment on your local repo. Here are additional changes that you should make to help with testing. If you do not make these edits, you might delete or mis-edit Hack for LA data, delete team members,  generate false or junk notifications, and/or create junk issues on HfLA's live project board. 
  - In the `schedule-monthly.yml` file:
    - Replace HfLA tokens with a personal token, 4 locations. 
    - Make sure your personal token has a different name than HfLA's token
    - Line 12, redirect to your personal repo.
  - In `contributors-data.js`:
    - Lines 16-21, replace current with:
```
let oneMonthAgo = new Date();                        // oneMonthAgo instantiated with date of "today"
oneMonthAgo.setMonth(oneMonthAgo.getMonth() - 4);   // please make sure that the digit is set to '4'
oneMonthAgo = oneMonthAgo.toISOString();
let twoMonthsAgo = new Date();                        // twoMonthsAgo instantiated with date of "today"
twoMonthsAgo.setMonth(twoMonthsAgo.getMonth() - 5);  // please make sure that the digit is set to '5'
twoMonthsAgo = twoMonthsAgo.toISOString();
```
 - In `create-new-issue.js`:
    - Line 14, change "owner" to your personal repo
    - Lines 36-37, replace current with:
```
  let removedList = removeList.map(x => "@ " + x).join("\n");  // please make sure there is a space after "@  "
  let notifiedList = notifyList.map(x => "@ " + x).join("\n");  // please make sure there is a space after "@  "
```
 - In `comment-issue.js`:
    - Line 12, change the issue number to an issue in your personal repo
    - Line 18, change "owner" to your personal repo

### Background
  - Referring to #4768 also: Previously, both `schedule-monthly.yml` and `schedule-thu-1100.yml` have had bugs and have not run correctly for over a year. Both actions were intended to run once a month, and both involved listing inactive members and then performing some action with the list. For these reasons, we decided that these two should be consolidated into a single action. 
  - 4768 started the consolidation by:
    - renaming the **_then-current_** "monthly" action as a temporary file named `schedule-monthly-PREV.yml`, then 
    - renaming the **_then-current_** "Thursday" action as the refactored `schedule-monthly.yml` (and deleting the now-empty `schedule-thu-1100.yml`)
  - 4788 finishes the consolidation by:
[...]

PR #5641, opened 2023-10-02
Add check for actions triggered by cron

aramattamara:
Improves #5639 

  - Added check for actions triggered by cron. 

  - #5639

### DIFF
--- a/.github/workflows/schedule-fri-0700.yml
+++ b/.github/workflows/schedule-fri-0700.yml
@@ -7,6 +7,7 @@ on:
 jobs:
   Add-Update-Label-to-Issues-Weekly:
     runs-on: ubuntu-latest
+    if: github.repository == 'hackforla/website'
     steps:
       - uses: actions/checkout@v3
       - uses: actions/github-script@v6
--- a/.github/workflows/schedule-fri-0700.yml
+++ b/.github/workflows/schedule-fri-0700.yml
@@ -7,7 +7,6 @@ on:
 jobs:
   Add-Update-Label-to-Issues-Weekly:
     runs-on: ubuntu-latest
-    if: github.repository == 'hackforla/website'
     steps:
       - uses: actions/checkout@v3
       - uses: actions/github-script@v6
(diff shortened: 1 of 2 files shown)


t-will-gillis (changes requested):
Hi @aramattamara  Thank you for the PR which appears to be a failsafe to prevent a GHA from running inadvertently when a dev has authorized a token in their personal repo with the same name as the HfLA token. I agree with you that this is a worthwhile and easy ‘safety’ measure.  

For the `schedule-monthly-PREV.yml` file, ideally this will be deleted by PR #5467 before it is scheduled to run again on Nov 1st, but just in case it isn’t, this is an easy solution. Incidentally, to be doubly sure lines 4 & 5 can be replaced with `workflow_dispatch:`. (Of course, I should have done this from the beginning but I did not think the last PR was going to be active this long.)  

For the `schedule-fri-0700.yml`, I don’t believe that the repo check is needed because the `secrets.IN_PROGRESS_COLUMN_ID` only refers to HfLA’s Project Board. Since the personal repos do not know this secret, they won’t trigger the action. For that reason, and since we have not tested to confirm that there are no unexpected results from adding the repo check, I do not believe we should update this GHA (If it ain’t broke…)  

For the future, this will be a good reminder to make sure that scopes are set appropriately to  prevent individuals from accidentally triggering GHAs- thank you for pointing this out.

If you could make the change(s) to `schedule-monthly-PREV.yml`, I can re-review and merge it. 

I will add labels, and if you think of any notes to add to the description for clarification, please do.

Thanks!

aramattamara:
Hi @t-will-gillis, thanks for your detailed feedback! I've implemented the changes you suggested in the `schedule-monthly-PREV.yml`. I've replaced lines 4 and 5 with `workflow_dispatch`.

I also reverted the `schedule-fri-0700.yml`, thanks.

t-will-gillis (approved):
Hey @aramattamara Thanks for the changes!

PR #5767, opened 2023-10-23

elizabethhonest:


roslynwythe (approved):
Using a file comparison extension in VS Code, I confirmed that there are no differences in _wins-data.json between update-wins-data and gh-pages, aside from extra whitespace in the Dean Church overview - VRMS project submission which will not affect the web page.

PR #6206, opened 2024-02-01
Updated section 2.7a information

jtw007:
Fixes #5697 

### What changes did you make?
  - updated git workflow instructions at section 2.7a

### Why did you make the changes (we will use this info to test)?
  - to keep information up to date

For Reviewers: Do not review changes locally, rather, review changes at https://github.com/jtw007/website/blob/update-contributing-section-2.7a/CONTRIBUTING.md 

### Screenshots of Proposed Changes Of The Website  (if any, please do not screen shot code changes)

<details>
<summary>Visuals before changes are applied</summary>

Screen Shot 2024-02-01 at 13 59 08

</details>

<details>
<summary>Visuals after changes are applied</summary>

Screen Shot 2024-02-01 at 13 59 18

</details>

PR #6217, opened 2024-02-03
Revert "Updated section 2.7a information"

t-will-gillis:
Reverts hackforla/website#6206


t-will-gillis:
I am cancelling this revert, and instead changing these (post-merge) observations to an ER #6218

Issue #5473, opened 2023-09-10
 Resolve CodeQL query alert 56


roslynwythe:
### Prerequisite
1. Be a member of Hack for LA. (There are no fees to join.) If you have not joined yet, please follow the steps on our [Getting Started page](https://www.hackforla.org/getting-started).
2. Before you claim or start working on an issue, please make sure you have read our [How to Contribute to Hack for LA Guide](https://github.com/hackforla/website/blob/7f0c132c96f71230b8935759e1f8711ccb340c0f/CONTRIBUTING.md).

### Overview
We need to analyze the query "Use of returnless function" which appears in the CodeQL code alert [56](https://github.com/hackforla/website/security/code-scanning/56) then either recommend dismissal of the alert or update code to resolve the alert. 

### Action Items
- [x] DO NOT DISMISS ANY ALERTS.  Dismissal of alerts should be done by dev leads only after review of the recommendation
- [x] Browse to the links in the following two Action Item and read the contents.  Click "See More" to view Recommendations, Examples and References.  
- [x] https://github.com/hackforla/website/security/code-scanning/56
- [x] In a comment in this issue, add your analysis and recommendations.  For each alert, the recommendation can be one of the following: `dismiss as test`, `dismiss as false positive`, `dismiss as won't fix`, or `update code`. 
- [x] If the recommendation is to update code:
   - [x] create an issue branch and proceed with the code update
   - [x] Use docker to test locally, ensuring that there are no changes to any affected webpage(s)
   - [x] proceed with pull request in the usual manner 
- [ ] If the recommendation is to dismiss, describe your reason for dismissal in the comment, then move the issue to `Questions/In Review` and apply the label `ready for dev lead`.   

### For merge team/dev lead
- [ ] If recommendation to dismiss is approved, dismiss the alert with a comment, then close the issue as completed.
- [ ] When this issue is closed please check it off (under "Issues") in #5159.  If all child issues are closed, close #5159 as completed.  
  
### Resources/Instructions
- [GitHub CodeQL documentation](https://codeql.github.com/docs/codeql-overview/about-codeql/)
- This issue is part of #5060

ajb176:
ETA: One week
Availability: Evenings

ajb176:
The code should be updated. 

Firstly, the same returnless function also occurs in the right-col-content.js file in the same directory, which is importing the same script from ./utility, so this and issue #6479 should be dealt with the same way. Because the issue has already been assigned, I'll just try to coordinate with them. 

[...]

PR #6536, opened 2024-03-27
refactor: use anonymous function as event listener callback for listing meetings

### LINKED ISSUE #6479: Resolve CodeQL query alert 57
### Prerequisite
1. Be a member of Hack for LA. (There are no fees to join.) If you have not joined yet, please follow the steps on our [Getting Started page](https://www.hackforla.org/getting-started).
2. Before you claim or start working on an issue, please make sure you have read our [How to Contribute to Hack for LA Guide](https://github.com/hackforla/website/blob/7f0c132c96f71230b8935759e1f8711ccb340c0f/CONTRIBUTING.md).

### Overview
We need to analyze the query "Use of returnless function" which appears in the CodeQL alert [57](https://github.com/hackforla/website/security/code-scanning/57) then either recommend dismissal of the alert or update code to resolve the alert. 

### Action Items
- [x] DO NOT DISMISS ANY ALERTS.  Dismissal of alerts should be done by dev leads only after review of the recommendation
- [x] Browse to the links in the following two Action Item and read the contents.  Click "See More" to view Recommendations, Examples and References.  
- [x] https://github.com/hackforla/website/security/code-scanning/57
- [x] In a comment in this issue, add your analysis and recommendations.  For each alert, the recommendation can be one of the following: `dismiss as test`, `dismiss as false positive`, `dismiss as won't fix`, or `update code`. 
- [x] If the recommendation is to update code:
   - [x] create an issue branch and proceed with the code update
   - [x] Use docker to test locally, ensuring that there are no changes to any affected webpage(s)
   - [x] proceed with pull request in the usual manner 
- [ ] If the recommendation is to dismiss, describe your reason for dismissal in the comment, then move the issue to `Questions/In Review` and apply the label `ready for dev lead`.   

### For merge team/dev lead
- [ ] If recommendation to dismiss is approved, dismiss the alert with a comment, then close the issue as completed.
- [ ] When this issue is closed please check it off (under "Issues") in #5159.  If all child issues are closed, close #5159 as completed.  
  
### Resources/Instructions
- [GitHub CodeQL documentation](https://codeql.github.com/docs/codeql-overview/about-codeql/)
- This issue is part of #5060

aqandrew:
Closes #6479 

### What changes did you make? / Why did you make the changes (we will use this info to test)?

See [this issue comment](https://github.com/hackforla/website/issues/6479#issuecomment-2023422712) for my analysis/recommendation.

<details>

https://github.com/hackforla/website/assets/9038965/2f0a91fd-bda6-4ece-b8db-183246dca94b

</details>

<details>

https://github.com/hackforla/website/assets/9038965/0d6e850a-5c04-4d30-971a-6a7c9f49fb5e

</details>

### DIFF
--- a/assets/js/right-col-content.js
+++ b/assets/js/right-col-content.js
@@ -8,10 +8,16 @@ import { getEventData, insertEventSchedule } from "./utility/api-events.js";
   const eventData = await getEventData();
 
   //Displays/Inserts event schedule to DOM
-  document.addEventListener(
-    "DOMContentLoaded",
-    () => { insertEventSchedule(eventData, "events"); }
-  );
+  if (document.readyState === "loading") {
+    document.addEventListener(
+      "DOMContentLoaded",
+      () => { insertEventSchedule(eventData, "events"); }
+    );
+  }
+  else {
+    insertEventSchedule(eventData, "events");
+  }
+
   //Displays/Inserts the user's time zone in the DOM
   document
     .querySelector("#userTimeZone")
--- a/assets/js/right-col-content.js
+++ b/assets/js/right-col-content.js
@@ -10,7 +10,7 @@ import { getEventData, insertEventSchedule } from "./utility/api-events.js";
   //Displays/Inserts event schedule to DOM
   document.addEventListener(
     "DOMContentLoaded",
-    insertEventSchedule(eventData, "events")
+    () => { insertEventSchedule(eventData, "events"); }
   );
   //Displays/Inserts the user's time zone in the DOM
   document


ajb176:
ETA: 24 hours

w1ld-r:
Review ETA: Wednesday 3/27 10pm CST

w1ld-r (changes requested):
Hey @aqandrew, 

Your implementation overall looks correct and I would think it would work, but the data doesn't appear to be loading properly after I spin up my Docker container and go to localhost:4000. This is what I'm seeing on your branch:

Screenshot 2024-03-27 at 3 10 29 PM

I dug a little deeper and I think the `insertEventSchedule` function being called before the DOMContentLoaded event has occurred and eventData is available, but we want it to be called after. 

The approach below worked for me. Can you try it and see if it works for you, too?

```
import { getEventData, insertEventSchedule } from "./utility/api-events.js";

/**
 * This type of function is called an IIFE function. The main function is the primarily controller that loads the recurring events on this page.
 * Refer: https://developer.mozilla.org/en-US/docs/Glossary/IIFE
 */
(async function main() {
  // Wait for DOMContentLoaded event
  await new Promise((resolve) => {
    document.addEventListener("DOMContentLoaded", resolve);
  });

  // Fetch event data
  const eventData = await getEventData();

  // Displays/Inserts event schedule to DOM
  insertEventSchedule(eventData, "events");

  // Displays/Inserts the user's time zone in the DOM
  document
    .querySelector("#userTimeZone")
    .insertAdjacentHTML("afterbegin", timeZoneText());
})();

```

This is what I'm seeing with the above adjustments:

Screenshot 2024-03-27 at 4 58 47 PM

ajb176 (changes requested):
I'm seeing the same issue as @gaylem with the schedule not populating, nice catch. 

I'm working on issue #5473 which faces the exact same issue with the same function on a different webpage. I looked through Andrew's analysis and we both came to the same conclusion that what was causing the error was passing a function with parameters to the event listener, which was causing the function to immediately run and try to pass its return value as the parameter. The function reference passed to addEventListener is supposed to be returnless, but because parameters were passed the EventListener assumes it's supposed to run the function and take its return value as its Listener. 

What you mentioned @gaylem as the possible reason for the issue actually seems to be the way the current live website is operating and has been operating for a long time. The function which was supposed to wait for the DOM content to load before firing was just immediately running. At least that's what I assumed, I could be wrong. Andrew's solution was meant to fix that by using an arrow function. Thanks for taking the time to write an alternate solution, I'll try it out and also try to dig into why Andrew's solution isn't working, because I was going to approach the problem in a similar way. Maybe an issue with the this context changing? 

I'll poke around and update because this issue and #5473 should be handled the same way.

ajb176:
Here's my best guess, which could be wrong:

The original code causing the CodeQL error wasn't designed correctly and only worked because it didn't do what it was intended to do. 

Because the script is asynchronous, it's possible that the event listener is set up after the DOMContentLoaded event, in which case it would be waiting for an event that already occurred and would never fire. In practice the live code incidentally works because the await operator loading the event data organically took longer than the DOM content loading, and once the script reached the event listener the function would immediately fire not because the condition was met but because it was loaded incorrectly to the event listener (with parameters). 

Because Gayle's code immediately sets up the event listener, it does so before the DOM content fully loads therefore it doesn't miss the DOMContentLoaded event. The code also works if you remove the await block and event listener altogether, in which case it functionally operates the same way as the live code. I would suggest making a slight change to make sure the event listener missing the event isn't possible. 

I have some code that seems to be working, I'll make a PR for #5473 in a couple hours and add you two as reviewers, because the issues are basically identical.

ajb176:
Please see [here](https://github.com/hackforla/website/pull/6543)

w1ld-r:
Hey @ajb176, good job on that other issue! I was curious so just tested your approach on this issue and I'm still not seeing the data render. This is the code I used:

```
import { getEventData, insertEventSchedule } from "./utility/api-events.js";

/**
 * This type of function is called an IIFE function. The main function is the primarily controller that loads the recurring events on this page.
 * Refer: https://developer.mozilla.org/en-US/docs/Glossary/IIFE
 */
(async function main() {
  const eventData = await getEventData();

  //Displays/Inserts event schedule to DOM
  document.addEventListener(
    "DOMContentLoaded",
    function () { insertEventSchedule(eventData, "events") }
  );
  //Displays/Inserts the user's time zone in the DOM
  document
    .querySelector("#userTimeZone")
    .insertAdjacentHTML("afterbegin", timeZoneText());
})();

```
What a mystery! The only explanation I can think of as to why your solution works on your ticket but not here is that this code is pulling a literal ton of data and parsing it all out so that it's formatted properly on the client. Perhaps this is causing enough of a lag that we need to handle the promises with more care in this instance? 

UPDATE: Just read your comment more closely and I think we're both saying basically the same thing lol. If there's a way to simplify my approach further, that works for me.

ajb176:
Hey @gaylem, thanks for the review. 

Could you re-check the code you quoted and what's being run, because it adds an anonymous function but doesn't have the if and else clauses to deal with the race condition of the DOM. 

I adapted the code from the other PR here and it seems to be working on my end: 

```
import { getEventData, insertEventSchedule } from "./utility/api-events.js";

/**
 * This type of function is called an IIFE function. The main function is the primarily controller that loads the recurring events on this page.
 * Refer: https://developer.mozilla.org/en-US/docs/Glossary/IIFE
 */
(async function main() {
  const eventData = await getEventData();

  //Displays/Inserts event schedule to DOM
  if (document.readyState==="loading") 
  {
  document.addEventListener(
    "DOMContentLoaded",
    function(){ insertEventSchedule(eventData, "events"); }
  );
  }
  else { insertEventSchedule(eventData, "events") }

  //Displays/Inserts the user's time zone in the DOM
  document
    .querySelector("#userTimeZone")
    .insertAdjacentHTML("afterbegin", timeZoneText());
})();
```

My code was based on yours, except I just added a check to make sure the event listener wasn't added too late, and an else clause to populate the schedule in case it was too late to catch the event.

aqandrew:
Thank you both for being so thorough!

I believe you're both correct about race conditions between DOMContentLoaded and `getEventData` returning. I checked my original code after hard reloading, and saw a blank right column like @gaylem was seeing.

@ajb176 your version of this function works as expected! Even after hard reloading. 🙂 I'll update my implementation to match yours. It feels a bit wrong to have the exact same control flow shared across 2 files--although I would be happy to merge this PR (after revision) and #6543, then open up a new issue to create a utility function that both right-col-content.js and project-meetings.js can share. Thoughts?

w1ld-r:
@ajb176 Ohhh got it, thanks for calling that out. My eyes must have been blurry. 

@aqandrew Glad you got it working! And I like where your head is at re: creating a utility function. If you want to create an ER I'd be happy to add my two cents if you want them!

ajb176 (approved):
Looks good to me! 

I'd suggest updating the original post with a line mentioning the race condition as a reason for making the changes, just so things are clear for the merge team.

Sharing the control flow for the same task would be a good idea, thanks for taking the initiative to make the ER.

w1ld-r (approved):
Hooray @aqandrew -- looks great!

* You linked the issue correctly in your PR description
* Your branch is correct
* You clearly described the changes, stated the reason for making them, and provided before/after screenshots
* The data is loading and looks good across mobile and desktop on Chrome.

Nicely done -- Approved! 👍

PR #6773, opened 2024-04-28
Auto response to feature branch issues - #1940

### LINKED ISSUE #1940: GitHub Actions: Automatic response to feature branch issues
### Overview
We need to create a GHA to post a comment on any issue that gets a label `feature: feature branch` with details about what the developer should do.  If the label gets removed, the message should be removed.  This should be a new job that runs in parallel with the other jobs under the "[Issue Trigger](https://github.com/hackforla/website/blob/gh-pages/.github/workflows/issue-trigger.yml)" action.

### Action Items
- [x] Trigger: On event `issues`: activity `opened` if the issue has the `feature: feature branch` OR activity `labeled` with the label `feature: feature branch`
    - [x] Add a comment to the issue:
```
Hi @[REPLACE WITH AUTHOR]
Based on the `feature: feature branch` label, this issue should target a feature branch.  Please consult the instructions on [working off of a feature branch](https://github.com/hackforla/website/wiki/How-to-work-off-of-a-feature-branch)
```
- [x] Trigger: On event `issues`: activity `unlabeled` - if the label `feature: feature branch` is removed
   - [x] Hide (as "outdated") the comment described above

### Checks

- [x] When the `feature: feature branch` label is added during issue creation, the comment is added
- [x] When the `feature: feature branch` label is applied after issue creation, the comment is added
- [x] When a **different label** is added during issue creation, the additional text is **not** added.
- [x] When the `feature: feature branch` label is removed, the comment is hidden
- [x] When a **different label** is removed, the comment is **not** hidden

### Resources/Instructions

#### Additional resources:
[actions/github-script](https://github.com/actions/github-script)
Bonnie test text to remove
Bonnie second test

partapparam:
Fixes #1940 

  - Updated `issue-trigger.yml` to include tracking of `labeled` and `unlabeled` events
    - When the `feature: feature branch` label is added during issue creation, or after issue creation, the comment is added
    - When the `feature: feature branch` label is removed, the comment is hidden
  - Added **feature-branch-comment** folder in `github-actions/trigger-issue`
    - Added `add-comments` script that will trigger on `labeled` event and add a comment to the issue
    - Added `outdate-comments` script that will trigger on `unlabeled` event and hide the comment as 'OUTDATED'
    - Added template file with comment to add 
  - Added `hide-issue-comment.js` in `github-actions/utils` folder
    - Will hide the comment and mark it as 'OUTDATED' 

To post a comment on any issue that gets a label `feature: feature branch` with details about what the developer should do. This was done in accordance with the below requirements.
  - When the `feature: feature branch` label is added during issue creation, the comment is added
  - When the `feature: feature branch` label is applied after issue creation, the comment is added
  - When a different label is added during issue creation, the additional text is not added.
  - When the `feature: feature branch` label is removed, the comment is hidden
  - When a different label is removed, the comment is not hidden

No visible changes to site


t-will-gillis (changes requested):
Hey @partapparam Great work and thanks for taking up this issue. Since this involves a GitHub Action, you will need some way to demonstrate the automation is working. The most direct way is to set up a testing environment in your own repo. Please see this update to [Hack for LA's GitHub Actions](https://github.com/hackforla/website/issues/6537#issuecomment-2041147335), in addition to the links on the original issue. Let me know if you need any additional info.

I was able to test this in my repo, and it appears to be mostly working. The main things that I notice are:
- for some reason that I was not able to track down quickly, on issue creation `issue-trigger.yml` attempts to run 4-5 times and skips all steps before it finally completes the initial steps of the workflow. This is a new behavior and does not happen currently.  

  <Details>
  Screenshot 2024-04-28 160016

  </Details>  
  
- I was testing adding and removing the `Feature: Feature Branch` label. The message was getting added consistently. When the label was removed, most*** if not all of the times the message was not hidden. ***Bringing this up as something to confirm, once or twice the message did not get hidden but this may due to my testing.

Other notes:
- The label referenced by the original issue is incorrect and should be `Feature: Feature Branch`. Please update all label references to match the actual label.
- On `issue-trigger.yml`, please add blank lines between jobs (around lines 103 and 88) to match the formatting of the rest of the file. 
- Similarly, please capitalize the first word of all new comments, and otherwise use lowercase to match the existing format and keep the code clean and consistent.
- This is a lesser item: I am thinking that it the file name could be more descriptive, for example `add-feature-branch-comment.js` and `get-feature-branch-comment.js` for the "outdate-comment.js" (especially since the "outdating" is not happening in this file) . 

Thanks again for taking on this issue

partapparam (commented):
@t-will-gillis thank you for the detailed comment. I updated the code with your feedback.

I dug into the GHA docs and found out why there were "Skipped" jobs. 

> If you specify multiple activity types, only one of those event activity types needs to occur to trigger your workflow. If multiple triggering event activity types for your workflow occur at the same time, multiple workflow runs will be triggered. For example, the following workflow triggers when an issue is opened or labeled. If an issue with two labels is opened, three workflow runs will start: one for the issue opened event and two for the two issue labeled events.

> For example, if you want the workflow to run when a specific label is added to an issue, you can trigger on the issues labeled event activity type and use a conditional to check what label triggered the workflow. The following workflow will run when any label is added to an issue in the workflow's repository, but the run_if_label_matches job will only execute if the label is named bug.

So by tracking the `Feature: Feature Branch` label, adding or removing any label will trigger the workflow, but it will not run the job due to the filter.

partapparam:
@t-will-gillis attached are photos of my test. I added test comments in the issue to ensure no other comments are marked `outdated`.

Comment marked as outdated.
Screenshot 2024-04-29 at 8 16 15 PM

The issue with multiple comments marked as outdated that were added for `Feature: Feature Branch` label.
Screenshot 2024-04-29 at 8 15 25 PM

This image shows that the workflow is triggered, but the job is skipped. 
Screenshot 2024-04-29 at 8 11 07 PM

t-will-gillis:
Hey @partapparam Thank you for your revisions and thanks for researching the reason for the multiple false starts on the workflow. 

I will try to look at this again in the next day or so.

t-will-gillis:
Hey @partapparam so first thing- this looks great and it looks like this PR does everything the initiating issue asked it to do. Well done!

About the workflow calls, I understand what you are saying and I was playing with your code a little bit to see what it would do. You already know this, but I don't see a way to get around this workflow being called every time an individual label is added or removed. On one hand it doesn't feel change things but on the other it is annoying to have 9-10 job runs just to open an issue and add labels. 

I apologize in advance, but I would like to discuss this effect with Bonnie and the team and ask if this is really what we want to happen and if this particular functionality truly is needed.  

Again, emphasize great work on your part

t-will-gillis (changes requested):
Hi @partapparam Discussed this PR with Bonnie and others at the Dev meeting earlier. Bonnie is OK with the 'nuisance' job runs created by the `labeled` and `unlabeled` triggers, so we can proceed.

Thanks for making the previous changes. Here are some additional items I noticed:
- `feature-branch-comment.md` on line 3, "branch" needs to be capitalized so that the `hide-feature-branch-comment.js` can find it
- in `issue-trigger.yml`, for both the "Add-Feature-Branch-Comment" and "Hide-Feature-Branch-Comment" steps:
  - On the if statements you can remove "github.event_name == 'issues' " because lines 2 and 3 cover this. So you can replace, for example:
 ` if: "${{ github.event_name == 'issues' && (github.event.action == 'labeled' && github.event.label.name == 'Feature: Feature Branch')}}" `
  with
  `if: "${{ github.event.action == 'labeled' && github.event.label.name == 'Feature: Feature Branch' }}"`
  - (In a future issue, this line could also be deleted from the steps above)
  - For both, the permissions are not needed, that is these lines can be removed in both steps:  
  
    ```
    permissions:
         issues: write  
    ```  

  - Lines 102/103 and 118/119 do not need to be returned and could be replaced with:  
  
    ```
    script({g: github, c: context})  
    ```  
  

  - this is minor but would you please capitalize "Hide" on line 113 +/-

Otherwise, I tested this with [one issue](https://github.com/t-will-gillis/website/issues/736) by adding and removing `Feature: Feature Branch`. I noticed that the "Hide" step only finds the last comment to close but this shouldn't matter (there shouldn't be multiple open feature branch comments). Also notice that when the comment is closed, the close message is a generic "This comment has been minimized"

Thank you again

partapparam:
@t-will-gillis thank you so much for all the work and time you're putting into this PR!!

I've made the requested changes:
- Removed `permissions`
- Removed `github.event_name == 'issues'`
- Capitalized `Hide`
- Removed `return` statement from both Jobs. 


Regarding the message `this comment has been minimized`. This seems to be a bug in the Github GraphQL API. See this attached discussion. [minimizeComment](https://github.com/orgs/community/discussions/19865#discussioncomment-7977097)

The `minimizeComment` is only available through the GraphQL API, not REST, and seems to be the only route to achieve the desired outcome. 

Testing the response from the GraphQL API, I was able to replicate the error mentioned in the link above. I pass in the `OUTDATED` flag as `minimizeReason`, but it returns as `outdated` from the API. I believe this is causing the generic message.

Screenshot 2024-05-07 at 2 38 14 PM

t-will-gillis (approved):
Hey @partapparam Fantastic- thanks for making the edits, and for the explanation regarding the "OUTDATED" comment. I tested this once more and everything appears to work as expected. 

Thank you for your patience and your contribution!

PR #6777, opened 2024-04-29
Fix `Our Events` Card Glitch After Toggling and Resizing Window 

### LINKED ISSUE #6775: Fix `Our Events` Card Glitch After Toggling and Resizing Window
### Overview
We need to fix the visual glitch of the _Our Events_ cards on event page for better website user experience.

### Details
It has been identified that a combination of:
- toggling the dropdown on the **Our Events** cards, followed by 
- resizing window between mobile and tablet/desktop view

can cause a visual glitch where contents of the card is lost. See complete steps to reproduce this issue below.

<details><summary><b>Complete Steps to Reproduce the Issue</b></summary>

1. make the browser window narrow enough (width <=767 in Chrome developer mode) so cards in "Our Events" section will be displayed stacked in a single column
2. Open Events Page at https://www.hackforla.org/events/
3. interact with the arrow by each card to fold/unfold cards
4. resize/widen browser window so cards appear side by side (width >767 in Chrome developer mode)
5. issue reproduced:
&nbsp; a. if card was folded before resizing, the content of the card will not display after resizing
&nbsp; b. if card was unfolded before resizing, there will be an extra symbol `∟` (`U+221F`) displayed next to the title

</details>

<details><summary><b>Visual Demo of the Glitch</b> </summary>

</details>

### Action Items
- [x] Update the CSS and/or JavaScript code as needed to fix this issue.

### Resources/Instructions
- This issue was created per #6687 
- [MDN Web Docs - Window: resize event](https://developer.mozilla.org/en-US/docs/Web/API/Window/resize_event)
- [Events Page](https://www.hackforla.org/events/)

#### Possible Direction
Preliminary research shows the following code might be related to this problem, as the following event listener is not properly registered:
https://github.com/hackforla/website/blob/42954259e5cffa7ef40add46541fbcfbff0e58a8/assets/js/events.js#L30-L44

From [MDN Web Docs - Window: resize event](https://developer.mozilla.org/en-US/docs/Web/API/Window/resize_event):
> In some earlier browsers it was possible to register resize event handlers on any HTML element. It is still possible to set onresize attributes or use [addEventListener()](https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/addEventListener) to set a handler on any element. **However, resize events are only fired on the [window](https://developer.mozilla.org/en-US/docs/Web/API/Window) object (i.e. returned by [document.defaultView](https://developer.mozilla.org/en-US/docs/Web/API/Document/defaultView)). Only handlers registered on the window object will receive resize events.**

tony1ee:
Fixes #6775 

  1. (7e94b92) changed where `eventListener` is registered from `.flex-page-card'` to `window`
  2. (4262f25) added a condition before `handleScreenResize()` collapse cards to check if they have been manually unfolded, and if so, leave them unfolded

  - change No.1 was made to address the main issue in [#6775](https://github.com/hackforla/website/issues/6775), making the eventListener properly registered (with the current way of registration it won't be triggered) so it can function as intended to reset the cards after resizing
  - change No.2 was made to address a issue that existed in the code before but [was only exposed after change No.1 is made](https://github.com/hackforla/website/issues/6775#issuecomment-2081630841) now that the function can be triggered: `handleScreenResize()` will reset (collapse) all cards every time the window size is `#bp-below-tablet` after `resize`. The problem is it cannot tell whether the **previous state before `resize` event is also `#bp-below-tablet`**. If the resize happens from a `#bp-below-tablet` width to another `#bp-below-tablet` width, the expected behavior is to leave the card as is, rather than resetting/collapsing it. By adding a condition to check if card was manually unfolded (by checking existence of `active` in class list), the unfolded cards will not be unexpectedly collapsed, therefore fixing the problem.
 

### Screen Recordings of Proposed Changes Of The Website

<details>
<summary>Visuals before change 1 is applied</summary>

notice how the card will be empty after resizing from width 750 to width 800

b1

</details>

<details>
<summary>Visuals after change 1 is applied</summary>

notice how the card content **remains** after resizing from width 750 to width 800
  
a1

</details>

<details>
<summary>Visuals after change 1 is applied but before change 2 is applied</summary>

notice the card collapse unexpectedly while adjusting the width (<=767 the entire duration)

b2

</details>

<details>
<summary>Visuals after both change 1 and change 2 are applied</summary>

expected behavior: the unfolded card remains unfolded while adjusting the width (<=767 the entire duration)
  
a2

</details>

### DIFF
--- a/assets/js/events.js
+++ b/assets/js/events.js
@@ -27,7 +27,7 @@
       }
     }
   }
-  document.querySelector('.flex-page-card').addEventListener('resize',handleScreenResize)
+  window.addEventListener('resize',handleScreenResize) 
   function handleScreenResize(){
     if(document.body.clientWidth>767){
       const columns = document.querySelectorAll('.mobile-dropdown');
--- a/assets/js/events.js
+++ b/assets/js/events.js
@@ -27,19 +27,24 @@
       }
     }
   }
-  window.addEventListener('resize',handleScreenResize) 
-  function handleScreenResize(){
-    if(document.body.clientWidth>767){
-      const columns = document.querySelectorAll('.mobile-dropdown');
-      for(let column of columns){
+  window.addEventListener('resize', handleScreenResize); 
+
+  /**
+   * Handles the screen resize event for event cards
+   * When the screen width is greater than 767 pixels (tablet/desktop), disables mobile dropdown, removes the arrow and display card content
+   * When the screen width is less than or equal to 767 pixels (mobile), enables cards mobile dropdown and fold the cards, unless already unfolded ('active').
+   */
+  function handleScreenResize() {
+    const columns = document.querySelectorAll('.mobile-dropdown');
+
+    if (document.body.clientWidth > 767) {
+      for(let column of columns) {
         column.style.display='block';
         column.previousElementSibling.classList.remove('active');
       }
-    }
-    else{
-      const columns = document.querySelectorAll('.mobile-dropdown');
-      for(let column of columns){
-        if (column.previousElementSibling.classList.contains('active')){
+    } else {
+      for (let column of columns) {
+        if (column.previousElementSibling.classList.contains('active')) {
           // when collapsing cards, skip the ones unfolded manually
           continue;
         }
--- a/assets/js/events.js
+++ b/assets/js/events.js
@@ -39,6 +39,10 @@
     else{
       const columns = document.querySelectorAll('.mobile-dropdown');
       for(let column of columns){
+        if (column.previousElementSibling.classList.contains('active')){
+          // when collapsing cards, skip the ones unfolded manually
+          continue;
+        }
         column.style.display='none';
     }
   }


w1ld-r (changes requested):
Hey @tony1ee,

Amazing work on this issue! It functions as expected for me locally. You also set up your branch and linked out to the issue correctly. Well done!

**One observation that probably doesn't matter:**
- I noticed that the drop-downs toggle shut when the when the screen is reduced from > 767 to < 767. This *might* cause user frustration if someone is reading the content and tries to shrink their screen. Could disorient them a bit. 
- It's such an edge case, though, and fixing that seems like it would be way more complicated than its worth in this instance. Besides, how would we know someone is reading the content? I think addressing this would require too much of a redesign and falls outside of the scope of this issue. Just thought I'd mention it in case anyone else felt the same way. 

**Otherwise, I only have nit-picky change requests:**
- This is optional, but if you'd like to clean up the code you modified to follow the ([AirBnB style guide](https://github.com/airbnb/javascript?tab=readme-ov-file#whitespace--before-blocks)) that would look pretty. I haven't heard if we follow a specific style guide but this is the one I see most commonly used.
- Given this feature is a bit finicky, would you mind adding some more comments to the event listener for future devs so they understand why it's built this way? JSDoc comments would also be rad, but you can do it however you want. 

Thanks, dude! 👍

w1ld-r:
**Note for other reviewers:** The `Add Pull Request Instructions` check is failing because of [#6778](https://github.com/hackforla/website/issues/6778). The changes were reverted, but its still not working for this issue. Here's how I set up my branch locally so you don't have to figure it out: 

```
git checkout -b tony1ee-fix-events-card-glitch-6775
git pull https://github.com/tony1ee/website.git fix-events-card-glitch-6775
```

**Note for merge team:** I'm guessing we can ignore that failed check and just squash and merge when this gets two approvals. Let me know if you disagree.

Thinking-Panda (dismissed):
@tony1ee , nice work with the issue. Code changes give expected result. Website works well on my local machine. I approve!

tony1ee:
@gaylem Thanks for your thorough review.

You were correct to point out the behavior of the card when first switched to/loaded mobile view. I am not sure what is the expected behavior in terms of default card display. I would consider it out of scope of this issue and suggest doing an issue involving UI/UX to address it further.

As for your 2nd concern, I pushed another commit with comments and some format adjustments. I tried my best to balance between maintainability and keeping the changes within issue scope. Please let me know your thoughts!

w1ld-r (approved):
@tony1ee I agree with your comments and the changes look great, approved!

Issue #6463, opened 2024-03-14
Create wiki page to guide developers how to manage CodeQL alerts


roslynwythe:
### Overview
We require a wiki page to guide developers how to manage and resolve CodeQL alerts

### Action Items
- [ ] We are currently in the process of moving the old wiki to the new [website-wiki repo](https://github.com/hackforla/website-wiki), so we will not be making any changes or additions to the old wiki at this time. Thus, we will be adding wiki content through a different process now. Read [How to Contribute to the Wiki](https://github.com/hackforla/website/wiki/How-to-Contribute-to-the-Wiki)
- [ ] Create a wiki page "How to manage CodeQL alerts" that covers the following topics:
   - [ ] Overview of code scanning and its implementation in code.yml workflow
   - [ ] Important elements of the code scanning alert page and details, including the concept of the tracking issue
      - [ ] How to use alert tags to determine if an alert represents a security risk
      - [ ] Example query to find closed alerts of a specific query type
   - [ ] Options for resolving alerts 
   - [ ] how to resolve specific alert query types
      - [ ] Unused variable, import, function or class (for sample resolution see https://github.com/hackforla/website/security/code-scanning/94)
      - [ ] Potentially unsafe external link (see sample resolved alert https://github.com/hackforla/website/security/code-scanning/3)
      - [ ] Malformed id attribute (see sample https://github.com/hackforla/website/security/code-scanning/25)
      - [ ] Missing variable declaration (for sample resolution see https://github.com/hackforla/website/security/code-scanning/49)
      - [ ] Use of returnless function
      - [ ] Syntax error (see sample https://github.com/hackforla/website/security/code-scanning/97 not resolved)
      - [ ] Inclusion of functionality from an untrusted source (see sample resolution https://github.com/hackforla/website/security/code-scanning/37)
      - [ ] Superfluous trailing arguments (for sample resolution see https://github.com/hackforla/website/security/code-scanning/35)

### After this issue is completed
- [ ] Release dependency on #5242

### Resources/Instructions

- This issue resulted from #5005

roslynwythe:
### Draft Wiki page: How to manage CodeQL alerts

### Overview of CodeQL scanning
- CodeQL scanning is implemented via the workflow `.github/workflows/codeql.yml` which is triggered by push, pull request on gh-pages, and scheduled weekly.   See #4886 for details
- The queries (tests) executed by CodeQL are grouped into query packs.  Our workflow specifies the `security-and-quality` query pack.  
[...]

Issue #7151, opened 2024-07-28
Research and Provide a Demo of Unit Testing for GHAs


roslynwythe:
### Dependency
- [x] #6454 

### Overview
We need to develop a sample set of unit tests against the JavaScript code in HfLA GitHub Actions, along with a new workflow that will invoke the tests in response to push and pull request triggers, in order to implement a GitHub PR check that will help ensure the quality of GHA code.  

### Action Items
- [ ] Select one of the following JS test frameworks:  Jest, Mocha and Jasmine.  Consider:
   - [ ] widely-used JS test framework suitable for invoking tests against JavaScript code within the HfLA GitHub Actions
- [ ] Provide a comment documenting the selection, installation and configuration of the test framework
- [ ] In a new git branch, create a folder `unit-tests-GHAs`
   - [ ] Add unit tests for at least one GitHub Action
- [ ] Create a new workflow `.github/workflows/unit-tests-GHAs` that will invoke all of the unit tests in the new folder and will display the results for each test and will return a result scuh that merge will be blocked if any test fails.  
- [ ] Test in your fork of the website repository 
- [ ] Create a PR with instructions for other developers to invoke the workflow from their forks.  

### Resources/Instructions
https://jasmine.github.io/
https://raygun.com/blog/javascript-unit-testing-frameworks/
https://www.browserstack.com/guide/jest-vs-mocha-vs-jasmine
[How to Test a GitHub Action with GitHub Actions](https://dev.to/cicirello/how-to-test-a-github-action-with-github-actions-2hag)
[Publish and view Mocha test results in Github Actions](https://medium.com/@ma11hewthomas/publish-and-view-mocha-test-results-in-github-actions-5ae74b59f4d6)
[Test workflow from Mocha examples](https://github.com/mochajs/mocha-examples/blob/main/.github/workflows/test.yml)
- This issue is a follow-up to #2249.  In particular see https://github.com/hackforla/website/issues/2249#issuecomment-2113750578

roslynwythe:
@t-will-gillis Your advice would be appreciated.  Is it a good idea to take on this issue now,  or should we wait until the current GHA changes settle down?  And, are there any particular unit tests that you would like to suggest, or general guidelines for selecting sample unit tests?  We can discuss at tomorrow's meeting if you would like.  

I tend to think that the task related to local testing using netkos/act should be in a separate issue that has a dependency on this one.  Do you agree?

HackforLABot:
[...]

PR #8411, opened 2025-11-07
Demo of frontend JS unit testing

### LINKED ISSUE #7152: Research and Provide a Demo of Unit Testing for front-end JavaScript code
### Overview
We need to select a strategy and a test framework for running unit tests against the front-end JavaScript code in the HfLA website, so that tests can be run first locally and then ultimately as a Pull Request check, in order to ensure the quality of JavaScript code updates.   In this issue several sample unit tests will be developed and will be run locally from the command line.    

### Action Items
- [x] Select one of the following JS test frameworks:  Jest, Mocha and Jasmine.  Consider:
   - [ ] widely-used JS test framework suitable for invoking tests against JavaScript code on a Jekyll site
   - [ ] Ideally tests could be run either locally or in the context of GitHub PR checkes
- [ ] Provide comments documenting the selection, installation and configuration of the test framework
   - [ ] In a new git branch, create a folder under the website root named `unittests-JSfrontend`
   - [ ] In the new folder create several unit tests.  Provide comments describing your reasoning for selection of unit tests.  
   - [ ] Invoke the tests from the command line
   - [ ] provide comments documenting the commands for running the tests
   - [ ] Submit a pull request, and provide links to the relevant comments that will be useful for reviewers.  

### Resources/Instructions
https://jasmine.github.io/
[How to Test a GitHub Action with GitHub Actions]
- This issue is a follow-up to #2249

ryanfkeller:
Fixes #7152

  - Selected Jest as the test framework
  - Added "hfla_test" container that installs Jest and test dependencies, then runs all tests
  - Added some minor refactor to two frontend JS files to extract pure-logic functions
  - Added a Jest transformer that renders Jekyll front-matter and Liquid into pure JS
  - Added two unit tests for pure-logic functions
  - Added two integration tests for DOM/Jekyll-based
  - Added GHA workflow to run the tests on PRs targeting gh-pages

#### Test Framework Selection: Jest
I landed on targeting Jest for this infrastructure. Compared to the other options I was considering (Mocha, Jasmine, Vitest), Jest still meets the "widely used" criteria the best -- see here: https://npmtrends.com/jasmine-vs-jest-vs-mocha-vs-vitest

Over the course of this development, I came to the conclusion that we could probably use any of these frameworks successfully -- but Jest has some nice features already built in like code coverage, function mocking, and assertions that some other frameworks don't, and would require more dependencies to include.

Vitest was probably the next best choice. I read that Expunge Assist is transitioning from Jest to Vitest due to some React dependency issues, and Vitest is growing in popularity. However, we don't have that dependency issue, Jest is more ubiquitous, and we aren't using Vite, so I think Jest is still a preferable choice for us.

#### Testing Environment
I opted to create an additional container in our docker_compose.yml so that every dev will have access to the same testing environment. When this container is launched, docker pulls down node:20-alpine, installs npm, jest, and some other libs I used in test, and then runs all Jest tests.

The other notable thing in the Jest "environment" setup is the transformer, `jekyll-js-transformer.js`. Our frontend JS scripts use frontmatter and LiquidJS statements that are rendered away in Jekyll, but that means we can't import them as-is into JS tests without hitting syntax issues. I debated just using the Jekyll build output, but it's slow and doesn't play nicely with CI, so I opted to leverage Jest's "transformer" functionality. This transformer runs on any files that match the pattern described in the `jest.config.js` file, then essentially performs a render of those files using the real Jekyll data in the repo. Fortunately, the CJS-syntax-transformer works on both CJS and ESM targets!

Not being able to dynamically adjust the data in the render was a pain point for me, but ultimately all the solutions I came up with to allow for injecting custom test data (rather than the actual data) were super complicated, fragile, and slow, so I abandoned that effort. I think time is better spent refactoring Liquid/etc. out of logic that needs to be tested to that degree.

#### Code Updates
There are two things that I think need to be done to the actual frontend JS scripts in order to make them more testable -- and I did some of that on my "examples"
1. Files that use ESM syntax should have the extension `.mjs`. Since we use both CJS and ESM in our frontmatter scripts, the test framework (and test writers!) need a way to know what syntax is being used in the target. 
2. Functions that needs to be unit tested needs to be exported by the source so they can be imported by the test. The way I went about this was by creating seperate `-util` files that contained the exported files, that the original source then imported, SO that I didn't change the public API of the original sources. However, we could also do something like a conditional export of those functions if we are in a "test mode" or something... so I'm very open to opinions on that. 

#### Tests
The scope of "unit testing" varies a lot depending on who you ask. I decided to break the potential scope of useful and relevant unit tests into two levels: 
* "Unit" Tests: the unit-under-test in each test is an individual function, and test files should contain all the tests for the relevant functions in a given source file. Functions in this category should ideally not interact with the DOM or have any state or global side-effects. I'd also probably prefer if they didn't have Liquid or Frontmatter since that adds some dependency outside the function itself.
* "Integration" Tests: the unit-under-test is a complete script as it would be loaded in an HTML file. Test files should focus on an individual source file, and (if needed) provide a simulated DOM and expect that the source will be translated with real Jekyll data. 

I provided two examples of Unit Tests: 
* `current-projects-utils.unit.test.js` to show an example of unit testing a set of fairly complicated functions from a CJS-syntax-script
* `vrms-events-utils.unit.test.mjs` to show basic testing of an ESM syntax unit script. 

I also created two examples of Integration Tests:
* `hamburger-nav.intg.test.js` to show testing a script that interfaces with the DOM heavily
* `vrms-events.intg.test.mjs` to show testing a script that depends heavily on Jekyll-injected content

I'm not ready to claim that these scripts are completely sufficient, but they did produce ~100% branch and statement coverage on the target files, so I think they are representative. 

#### Running Tests as a Developer
The tests run automatically when you launch the docker container. This can be done with the following command: 
```bash
docker compose up hfla_test
```

<details>
<summary>Expand for a screenshot of a test run. (Darn that last branch I can't test!)</summary>

image

</details>

#### Running Tests as CI
In addition to the developer environment, I've added the Jest call to the workflow that triggers on PR opening and updates targeting gh-pages, so the tests will run in those cases as well. The workflow steps are basically identical to the steps in setting up the new docker container -- install npm, install dependencies, run the tests. 

The tests are run on the source branch rather than the target branch, so I *hope* they run on this PR!

#### Currently Known Gotchas
* With the current configuration, we need to pull down node:20-alpine and install the dependencies for each time you want to re-run tests. There is some caching that is done, but it still feels like a lot of unnecessary work. It would probably be easier to just have a step where we build a docker image from node:20-alpine where we install our dependencies, then the container just calls that image... but that felt out of scope. Still, I think there's room for optimization here
* Mapping coverage "Uncovered Line #s" to the actual file lines is a little wonky. Jest is reporting the line numbers from the **transformed file**, which in this setup is not ever actually visible to the user. You have to do some reasoning to match line-numbers with source file line numbers. This could probably be helped by storing the transformed files somewhere accessible to users, OR by creating the text modification map and handing that to Jest, but for now that's the situation

#### Summary
I'm VERY interested in differing opinions or thoughts on how this can be better. I'm not married to really anything I've done here, and I think there are several ways it can be done differently -- and I'm super happy to change or re-do things. I'd also understand if this PR can't be merged as-is because it has several not-necessarily-trivial changes to source code.

All that said, I think this framework is pretty functional and I'm stoked to share it. I think that having some tests in place will make things like refactoring a lot easier, since we'll have the behavior already codified and an instant way to see success/failure. (I'm already eyeing that `createFilter` function in `current-project-utils.unit.test.js`).

Thanks for taking a look.

<h3>CodeQL Alerts</h3>

After the PR has been submitted and the resulting GitHub actions/checks have been completed, developers should check the PR for CodeQL alert annotations.

<details><summary>Check the PR's comments. If present on your PR, the CodeQL alert looks similar as shown</summary>
  
Screenshot 2024-10-28 154514

</details>

Please let us know that you have checked for CodeQL alerts. **Please do not dismiss alerts.**
- [ ] I have checked this PR for CodeQL alerts and none were found.
- [x] I found CodeQL alert(s), and (select one):
   - [x] I have resolved the CodeQL alert(s) as noted
   - [x] I believe the CodeQL alert(s) is a false positive (Merge Team will evaluate)
   - [ ] I have followed the Instructions below, but I am still stuck (Merge Team will evaluate)

<details><summary>Instructions for resolving CodeQL alerts</summary>

If CodeQL alert/annotations appear, refer to [How to Resolve CodeQL alerts](https://github.com/hackforla/website/issues/6463#issuecomment-2002573270).  

In general, CodeQL alerts should be resolved prior to PR reviews and merging

</details>

### Screenshots of Proposed Changes To The Website (if any, please do not include screenshots of code changes)
 
- No visual changes to the website


ryanfkeller:
One additional comment -- setting up the framework to enable the demos required in the issue was pretty complicated. I think that this issue was quite a bit bigger than 2pts. For reference, the research, experimentation, and implementation of this PR took me somewhere around 30-40h. I'd say at least 20h of that was strictly necessary. The sister task [#7151](https://github.com/hackforla/website/issues/7151) is likely to be similarly challenging and should maybe be scaled up.

daras-cu:
@ryanfkeller I updated the original issue size and complexity to reflect the actual time it took, thanks for bringing that up. Looks like Will is assigned to #7151 so hopefully he'll have a head start from looking at your work here.

ryanfkeller:
Plugging away at this a little bit. After some discussion last year, I'm agreed that the "integration" tests were out of scope, so I've kept only the unit tests. Additionally, I've re-consolidated the utils scripts, as breaking out the unit-testable functions into different files is more overhead than is appropriate.

ryanfkeller:
Hi @daras-cu & @t-will-gillis, after a lot of thought, I don't think this PR is the right approach for testing our frontend JS, and I'd like to close it.

The main issue is that any unit testing solution requires a complex transformation step to convert Jekyll/Liquid syntax into pure JS before the test runner can see it. This framework works, but it's brittle and I think it would be a significant barrier for new contributors trying to understand or extend the tests.

Beyond the infrastructure issue, I also found that the codebase itself isn't well-suited to unit testing. Very few functions are pure logic, and most touch the DOM directly, which requires a lot of setup to fake correctly. Tests end up being fragile and tightly coupled to implementation details.

That said, I think that we would actually benefit a lot from having something more like end-to-end testing of the rendered site. I explored the idea and put together a "test" PR of the concept here: https://github.com/ryanfkeller/website/pull/1171. I think this would be a good fit for the project and I'd love your thoughts on pursuing it!

